PR #138, opened 2019-05-09
BUG: SPEC filewriter scan numbering and newfile() when file exists

prjemian:
resolve problems with `newfile()` when file exists


prjemian:
coveralls reports a big drop in number of lines tested in filewriters.py.  This reduction is a false observation since the number of tests has increased from 1 to 8.  What it *means* is that there is more code to be tested.  The coveralls web page highlights in red what has not been tested.

tl;dr: acceptable as-is

PR #274, opened 2020-01-31
APS user operations could be in mode 1 OR 2

### LINKED ISSUE #273: APS has user beam when in mode: Bm Ln Studies: change inUserOperations()
APS is now in a mode (`Bm Ln Studies`) that allows user operations but is not `USER OPERATIONS`.  This mode should be added to `devices.ApsMachineParametersDevice.inUserOperations()` as an allowed mode.

prjemian:
fixes #273

### DIFF
--- a/apstools/devices.py
+++ b/apstools/devices.py
@@ -285,12 +285,10 @@ class ApsMachineParametersDevice(Device):
                 pass
 
         """
-        valid_modes = (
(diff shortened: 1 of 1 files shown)


prjemian:
travis-ci is failing on py3.7 due to ophyd version 1.4.0rc3, ignore that for now since that release is older

PR #277, opened 2020-01-31
replace .value with .get() 

### LINKED ISSUE #270: prefer EpicsSignal.get() over EpicsSignal.value for most recent value
We have [noticed](https://github.com/aps-8id-trr/ipython-8idiuser/issues/119) this change makes a difference.  Also in https://github.com/aps-8id-trr/ipython-8idiuser/issues/118.

Check the shutter support and the ProcessController support.  Others?

### LINKED ISSUE #272: ophyd does not camonitor, replace .value with .get()
Our assumptions when this was coded was that ophyd placed EPICS Channel Access monitors on all EpicsSignal objects, this `.value` is used to get the value of the PV.  Turns out this was an initial design of ophyd to use CA `get()` throughout the code.  The value reported by the `.value` property is from Python's memory and may not reflect the current value in the EPICS IOC.

We are running into recurrent problems that reveal this assumption, python makes decision based on stale PV values.

Refactor this code base to use `.get()` in cases where `.value` is used.

prjemian:
fixes #272 which is duplicate of and will fix #270

### DIFF
--- a/apstools/devices.py
+++ b/apstools/devices.py
@@ -166,7 +166,7 @@ def use_EPICS_scaler_channels(scaler):
         configuration_attrs = []
         for ch in scaler.channels.component_names:
             nm_pv = scaler.channels.__getattribute__(ch)
-            if nm_pv is not None and len(nm_pv.chname.value.strip()) > 0:
+            if nm_pv is not None and len(nm_pv.chname.get().strip()) > 0:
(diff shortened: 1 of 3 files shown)


prjemian:
passes unit tests locally and in travis-ci for py3.6 but py3.7 is failing for some other reason

PR #384, opened 2020-07-22
apsbss: print, not log from commands

### LINKED ISSUE #383: apsbss: some logging should revert back to print
Since #359, some of the print statements were converted to logger and these should have remained as print since they deliver expected output.  Check all the `cmd_` functions, for example.

prjemian:
fixes #383


prjemian:
Note: commit hash c9d9741 is for #383

Issue #430, opened 2020-09-28
New default branch


prjemian:
It is proposed to move the default branch from `master` to `main`. See https://github.com/github/renaming for GitHub's plan on this. This would require any and all checkouts of this repository to change the default local/remote branch to `main` and `<upstream/main>`.

This change will not change the API or break existing code so it is neither a major nor minor change, just a [patch](https://semver.org/) change.  It can be part of the next patch release (1.3.8).

prjemian:
Rename of default branch will also need to be configured on cloud support integrations such as travis-ci, RTD, Codacy, LGTM, documentation-generator, ... (whichever this repo uses).

prjemian:
This is a good HOW-TO guide (and WHY-TO as well): http://www.kapwing.com/blog/how-to-rename-your-master-branch-to-main-in-git/

Search found even more guidance: https://duckduckgo.com/?t=ffsb&q=how+to+rename+the+git+master+branch+to+main%3F&ia=web

prjemian:
I'm adding this to the milestone for release 1.3.10 expected for release by the end of 2020.

prjemian:
On Github:  

1. Create new `main` branch
2. Set new `main` branch as default in repo's Settings
3. Set base (target branch) of each open PR to `main` branch
  - short video: https://github.blog/2016-08-15-change-the-base-branch-of-a-pull-request
  - documentation: https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request
4. Delete the old `master` branch

When each remote is *pulled*, each will need to:
1. `git checkout main`
2. `git branch -d master`

PR #451, opened 2020-11-04
Undulator and Kohzu monochromator functionalities

gfabbris:
##### `ApsUndulator`

-  Adds some `Signal` components that are helpful in moving the undulator (as far as I know these don't exist in the EPICS support).

- Interactive `.undulator_setup` - this is certainly useful for 4-ID-D, not sure others would want it.

##### `KohzuSeqCtl_Monochromator`

- Adds `.calibrate_energy`.

### DIFF
--- a/apstools/devices.py
+++ b/apstools/devices.py
@@ -911,9 +911,9 @@ class TrackingSignal(Signal):
         ------
         ValueError
         """
-        if type(value) != bool:
-            msg = 'tracking is boolean, it can only be True or False.'
-            raise ValueError(msg)
+        if not isinstance(value, bool):
+            raise ValueError(
+                'tracking is boolean, it can only be True or False.')
 
 
 class ApsUndulator(Device):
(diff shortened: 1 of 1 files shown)


apstools/devices.py
```diff
@@ -897,6 +897,25 @@ def state(self):
...
+
+    def check_value(self, value):
+        """
+        Check if the value is a boolean.
+
+        Raises
+        ------
+        ValueError
+        """
+        if type(value) != bool:
```
prjemian, line 914: ```if not isinstance(value, bool):```

apstools/devices.py
```diff
@@ -897,6 +897,25 @@ def state(self):
...
+            msg = 'tracking is boolean, it can only be True or False.'
```
prjemian, line 915: It is more efficient when the message is created within the raised exception.  Only when `raise` is called will the message text be created.  So:

```
raise ValueError("tracking is boolean, it can only be True or False.")
```

apstools/devices.py
```diff
@@ -906,27 +925,65 @@ class ApsUndulator(Device):
...
+    location = Component(EpicsSignalRO, "Location", kind='config')
+    version = Component(EpicsSignalRO, "Version", kind='config')
+
+    # Useful undulator parameters that are not EPICS PVs.
+    energy_deadband = Component(Signal, value=0.0, kind='config')
+    energy_backlash = Component(Signal, value=0.0, kind='config')
+    energy_offset = Component(Signal, value=0, kind='config')
+    tracking = Component(TrackingSignal, value=False, kind='config')
+
+    def undulator_setup(self):
```
prjemian, line 960: I'd rather not add any code to `apstools` that involves command line Q&A such as this.  Can the same function be replaced with [getter/setter decorators](https://stackabuse.com/the-python-property-decorator/)?

prjemian (changes requested):
changes as noted

apstools/devices.py
gfabbris: I don't think this function could be replaced with getter/setter. I'll remove it.

prjemian:
Per #430, I'd like to change the default branch name from `master` to `main` done by December 10 as part of the milestone for the 1.3.10 release.  If this PR is still open at that time, you'll need to re-target this PR to point to the new branch, once that has been created.  Will proceed unless I hear comments.

prjemian (approved):
LGTM

prjemian:
@gfabbris Thanks!

PR #474, opened 2021-01-13
Help listruns() default to the current catalog in use.

### LINKED ISSUE #468: listruns() should get db from session if possible
When using an intake-style catalog, the `listruns()` defaults through to an existing `mongodb_config` instead.  Make this smarter (but not at the expense of more fragile).

prjemian:
FIX #468


prjemian:
Raising the priority since we are seeing this as a problem with some beamlines using the intake-style catalogs.

PR #497, opened 2021-03-02
add findpv(PVNAME) utility function

### LINKED ISSUE #496: What ophyd signal uses this EPICS PV?
This is a common question as an instrument package begins to grow.  Users know the EPICS PV but forgot which EPICS signal and then where is it defined.

prjemian:
Add `utils.findpv()` to fix #496


prjemian:
@gfabbris : Ready for review

apstools/utils.py
```diff
@@ -2325,3 +2327,203 @@ def copy_filtered_catalog(source_cat, target_cat, query=None):
...
+        Namespace dictionary of Python objects.
+
+    RETURNS
+
+    str or ``None``:
+        Name of the ophyd object.
+
+    EXAMPLE::
+
+        In [45]: findpv("adsimdet_cam_acquire")
```
gfabbris, line 2488: typo? should it be `findname`?

gfabbris (approved):
Looks great!!

apstools/utils.py
prjemian: yup, that's a copy/paste error caught in its native habitat

PR #504, opened 2021-03-22
overhaul of listruns() using pandas

### LINKED ISSUE #500: listruns() sort: improvements
While these changes in #498 fix the *reported* problem, the sort by start time might be improved:

* make it faster by having databroker do the sort
* provide sort key (and direction) as kwargs

What about sort keys that are *NOT* in the `start` document?

_Originally posted by @prjemian in https://github.com/BCDA-APS/apstools/issues/498#issuecomment-798665200_

prjemian:
fix #500 

Thanks to @strempfer, @rodolakis, and John Freeland for the suggestions!


prjemian:
CI is failing for py3.6 with `E   ModuleNotFoundError: No module named 'dataclasses'`.  That's because that was added in py3.7.  Time to drop py3.6.

apstools/utils.py
```diff
@@ -384,7 +393,377 @@ def itemizer(fmt, items):
...
+
+    def _check_cat(self):
+        if self.cat is None:
+            self.cat = getCatalog()
+
+    def _apply_search_filters(self):
+        since = self.since or "1995-01-01"
+        until = self.until or "2100-12-31"
+        self._check_cat()
+        cat = self.cat.v2.search(
```
gfabbris, line 560: one comment here...

One can actually merge the TimeRange and "regular" queries into something like:

```python
time_query = databroker.queries.TimeRange(since=since, until=until).query
merged_query = self.query.update(time_query) if self.query else time_query
cat = self.cat.v2.search(merged_query)
````

Now, it seems like this is rather useless for a MongoDB because it's very fast anyway:

```python
In [38]: from databroker import catalog; db = catalog['mongodb_config']

In [39]: %%time
    ...: time_query = TimeRange(until='2021-01-01')
    ...: query = dict(plan_name="scan")
    ...: query.update(time_query.query)
    ...: search1 = db.search(query)
    ...: 
    ...: 
CPU times: user 414 µs, sys: 47 µs, total: 461 µs
Wall time: 467 µs

In [40]: from databroker import catalog; db = catalog['mongodb_config']

In [41]: %%time
    ...: time_query = TimeRange(until='2021-01-01')
    ...: query = dict(plan_name="scan")
    ...: search2 = db.search(time_query).search(query)
    ...: 
    ...: 
CPU times: user 675 µs, sys: 0 ns, total: 675 µs
Wall time: 683 µs
```

but it does make more of a difference if the `db` is backed by files (using a "unpacked" `db` in my laptop with 1042 runs):

```python
In [11]: from databroker import catalog; db = catalog['4idd-nov20']

In [12]: %%time
    ...: time_query = TimeRange(until='2020-12-03')
    ...: query = dict(plan_name="scan")
    ...: query.update(time_query.query)
    ...: search1 = db.search(query)
    ...: 
    ...: 
CPU times: user 6.75 s, sys: 2.4 s, total: 9.15 s
Wall time: 12.6 s

In [13]: from databroker import catalog; db = catalog['4idd-nov20']

In [14]: %%time
    ...: time_query = TimeRange(until='2020-12-03')
    ...: query = dict(plan_name="scan")
    ...: search2 = db.search(time_query).search(query)
    ...: 
    ...: 
CPU times: user 13.2 s, sys: 4.69 s, total: 17.9 s
Wall time: 25.1 s
```

apstools/utils.py
```diff
@@ -384,7 +393,377 @@ def itemizer(fmt, items):
...
+    num: int = 20
+    reverse: bool = True
+    since: str = None
+    sortby: str = "time"
+    timefmt: str = "%Y-%m-%d %H:%M:%S"
+    until: str = None
+
+    _default_keys = "scan_id time plan_name detectors"
+
+    def _get_by_key(self, md, key):
```
gfabbris, line 508: Should this always search both "start" and "stop" docs (reporting missing if not there)? In one hand, it is simpler to handle it as is, and the user can force searching the "stop" doc anyway. But in the other hand, this assumes that for most metadata, only one of these docs will matter. That is true today, with only "time" as an exception.

apstools/utils.py
```diff
@@ -384,7 +393,377 @@ def itemizer(fmt, items):
...
+    def _apply_search_filters(self):
+        since = self.since or "1995-01-01"
+        until = self.until or "2100-12-31"
+        self._check_cat()
+        cat = self.cat.v2.search(
+            databroker.queries.TimeRange(since=since, until=until)
+        ).search(self.query or {})
+        return cat
+
+    def parse_runs(self):
```
gfabbris, line 565: It will be very useful to have the option to get the metadata for particular runs instead of the last few. So should we add an option here to ask for specific runs (by `uid` or `scan_id`)?

gfabbris (approved):
Looks good.

apstools/utils.py
prjemian: Thanks for the analysis!

apstools/utils.py
prjemian: Considering how to lower the barrier to use for those who do not read documentation.

apstools/utils.py
prjemian: Since I want to release in the next few days and this could take a bit longer to implement well, I'm moving this comment to be a new issue for the next release.

PR #514, opened 2021-06-03
refactor .value to .get()

### LINKED ISSUE #513: sscan support uses .value -- should use .get()
Discovered while setting up 11-BM-B

Look for same in other modules, as well.

prjemian:
FIX #513


prjemian:
`apstools.devices.ProcessController.value` needs special examination.

PR #531, opened 2021-09-03
Add PVPositionerSoftDone

gfabbris:
The `PVPositionerSoftDone` is basically the same as `ophyd.PVPositioner` that uses a soft (instead of EPICS) done signal. A few other (current) differences:
- There is an option to to initialize it with the readback and setpoint PVs:
```python
example = PVPositionerSoftDone("prefixPV:", readback_pv="myreadbackPV", setpoint_pv="mysetpointPV", name="example")
```
where, for instance, the readback PV is "prefixPV:myreadbackPV". The advantage of using this is that the readback/setpoint signals are already marked as hinted/normal and forces auto monitor and PV completion. Alternatively, you can still do something like:
```python
class MyPVPositionerSoftDone(PVPositionerSoftDone):
    readback = Component(EpicsSignal, "myreadbackPV")
    setpoint = Component(EpicsSignal, "mysetpointPV")

example = MyPVPositionerSoftDone("prefixPV:", name="example")
```

- The tolerance can be set on startup or after:
```python
example = PVPositionerSoftDone(
    "prefixPV:",
    readback_pv="myreadbackPV",
    setpoint_pv="mysetpointPV",
    tolerance = 0.1,
    name="example")
```
or
```python
example = PVPositionerSoftDone(
    "prefixPV:",
    readback_pv="myreadbackPV",
    setpoint_pv="mysetpointPV",
    name="example")
example.tolerance.put(0.1)
```
If not set, it will use `10^(-1*precision)`, where the setpoint precision is used.

- For certain EPICS devices (like the lakeshore 340) if a ramp rate is used, the setpoint signal is slowly changed, which the `PVPositionerSoftDone` will recognize as a "done" motion before it's truly done. To fix this, you can define a dummy `target` signal:
```python
class MyPVPositionerSoftDone(PVPositionerSoftDone):
    target = Component(Signal, value = 0, kind="omitted")

example = MyPVPositionerSoftDone(
    "prefixPV:",
    readback_pv="myreadbackPV",
    setpoint_pv="mysetpointPV",
    target_attr = "target",
    name="example"
)
```

PR #535, opened 2021-09-20
remove target_attr, keep target signal

prjemian:
refactor `target_attr` and `_target` into `target`, end user could override with named EpicsSignal if desired.

### DIFF
--- a/apstools/_devices/positioner_soft_done.py
+++ b/apstools/_devices/positioner_soft_done.py
@@ -8,6 +8,7 @@ from ophyd import PVPositioner
 from ophyd import Signal
 from ophyd import EpicsSignal
 from ophyd import EpicsSignalRO
+from ophyd.signal import EpicsSignalBase
 import logging
 
 
@@ -37,10 +38,6 @@ class PVPositionerSoftDone(PVPositioner):
 
         Defaults to ``10^(-1*precision)``,
         where ``precision = setpoint.precision``.
-    target_attr : str, optional
-        Used if the setpoint is controlled incrementally by EPICS
-        (like with a ramp). Then a target attribute signal must be
-        defined, and its name passed in this variable.
     kwargs :
         Passed to `ophyd.PVPositioner`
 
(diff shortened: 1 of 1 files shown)


prjemian:
@gfabbris This is ready for your review.

prjemian:
@gfabbris This is the most important to review (for you) on this project at this time.

apstools/_devices/positioner_soft_done.py
```diff
@@ -84,7 +95,7 @@ def cb_readback(self, *args, **kwargs):
         """
         Called when readback changes (EPICS CA monitor event).
         """
-        diff = self.readback.get() - self._target.get()
+        diff = self.readback.get() - self.setpoint.get()
```
gfabbris, line 98: This won't work well for some devices in which the setpoint changes when a ramp is used. For example, this is the case for the lakeshore 340. So, it should be ```diff = self.readback.get() - self.target.get()```

gfabbris (changes requested):
Looks good to me, only one issue.

prjemian:
Will need to resolve merge conflicts with upstream branch due to [new content in main branch](https://github.com/BCDA-APS/apstools/pull/531#issuecomment-933142013).

apstools/_devices/positioner_soft_done.py
prjemian: The `setpoint` is the ultimate temperature where the controller is headed.  The `_target` is the interim required value for the temperature control loop.  (For example, a controller's PID loop will ramp `_target` as the immediate objective to calculate the following error: `readback - _target`.)  This `_target` value may be changed incrementally (ramped) by the controller for a controlled transition from current to `setpoint`.  We do not need to examine the controller's interim `_target` to determine if the controller has reached the requested temperature.

Definitely: `done = |readback - setpoint| <= tolerance`

gfabbris: Currently, the Lakeshore 340 EPICS supports ramps the setpoint PV. Thus I think it's not great to call it setpoint in EPICS and target in Bluesky.

prjemian: You're saying the EPICS Lakeshore support uses `setpoint` for the interim setpoint of the ramp?
Doesn't the Lakeshore have a PV that reports where it is headed?  That's the PV to be called `setpoint` in Bluesky to be consistent with how it is used in other positioners, such as EpicsMotor.

The PTC10 has both: a setpoint where the controller is headed (that is `setpoint` in bluesky) and a setpoint for the PID loop (that would `target` in bluesky but this interim goal is not interesting to the user).

An interesting case is when `|target - setpoint| <= tolerance`, that's when we expect to be close enough that `|readback- setpoint| <= tolerance` could become `True`.

prjemian:
Upstream changes (from `main` branch) did not cause merge conflicts.  Good.

apstools/_devices/positioner_soft_done.py
prjemian: We would use this for more controllers than the Lakeshore?

prjemian:
@gfabbris Ready for your review again.

PR #543, opened 2021-10-01
Add device support for SRS570 from synApps

### LINKED ISSUE #541: SRS570 amplifier
Support the SRS570 amplifier.

Is the [preamps support at APS 8IDI](https://github.com/aps-8id-dys/ipython-8idiuser/blob/1198f3131cf971294bea254c1f298416c9fdbbdd/profile_bluesky/startup/instrument/devices/preamps.py#L14-L56) a starting point to be hoisted?

prjemian:
- FIX #541

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,7 +7,13 @@ Change History
 The project `milestones <https://github.com/BCDA-APS/apstools/milestones>`_
 describe the future plans.
 
-.. :1.5.3:  release expected by 2021-10-31
+.. :1.6.0:  release expected by 2021-02-01
+
+:1.5.3:  release expected by 2021-11-01
+
+   * `#541 <https://github.com/BCDA-APS/apstools/pull/541>`_
+      Add device support for SRS570 preamplifier from synApps.
+      Hoisted from 4-ID-C Polar and 8-ID-I XPCS.
 
 :1.5.2:  released 2021-09-29
 
--- a/apstools/synApps/__init__.py
+++ b/apstools/synApps/__init__.py
@@ -33,6 +33,7 @@ https://github.com/klauer/recordwhat
 # -----------------------------------------------------------------------------
 
 
+# structures
 from .asyn import *
 from .busy import *
 from .calcout import *
@@ -43,6 +44,9 @@ from .sscan import *
 from .swait import *
 from .transform import *
 
+# specific devices
+from .srs570_preamplifier import *
+
 __all__ = """
     AsynRecord
     BusyRecord
@@ -51,6 +55,7 @@ __all__ = """
     CalcoutRecordChannel
     EpidRecord
     SaveData
+    SRS570_PreAmplifier
     SscanRecord
     SscanDevice
     SwaitRecord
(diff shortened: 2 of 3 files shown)


prjemian:
Compare with other preamplifiers and generalize: Keithley 428 and Femto (USAXS has both).

Might be suitable as `ophyd.PVPositioner` **_if_** we can control the values for the setpoint to match the values the readback can provide.  This is controller-dependent.

gfabbris:
@prjemian: I double checked all the "value" PVs in the SRS570, and you either get the state number or a string with the value.

prjemian:
I'm seeing the same, thinking that is a legacy misdirection.  Numerical values should be numbers, not strings.  Can they be forced into numbers?  Not directly, as shown:

```python
(bluesky_2021_2) prjemian@zap:~/.../BCDA-APS/apstools$ ipython
Python 3.8.10 (default, May 19 2021, 18:05:58) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.22.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import ophyd

In [2]: !caput gp:gp:text1 "1.0"
Old : gp:gp:text1                    
New : gp:gp:text1                    1.0

In [3]: sens = ophyd.EpicsSignal("gp:gp:text1", name="sens", string=False)

In [4]: type(sens.get())
Out[4]: str

```

prjemian:
Can we subclass to force these to numbers?  thinking: `EpicsSignalNumber()`

prjemian:
On closer examination, `sens_num` is defined as a `mbbo` EPICS record type.  There is a confusing overlap between the enumeration value and the _sensitivity_ number which can take these values: `1 2 5 10 20 50 100 200 500`

The mapping between mbbo index of `5` (sensitivity value of `50`) and a sensitivity value of `5` (index of `2`) is an ambiguity that cannot be resolved unless string values for the sensitivity are defined.

prjemian:
## Conclusion

To make this a subclass of `ophyd.PVPositioner`, would need to have _gain_ (for setpoint and readback) be a number, not a string.  Is it really worth it?

prjemian:
Since other synApps hardware device support is located under `_devices`, relocate this new support with the others.

prjemian:
Also, let [*pint*](https://pint.readthedocs.io/en/stable/tutorial.html) handle the units conversions.  Example:

```python
In [57]: for units in amp.sensitivity_unit.enum_strs:
    ...:     print(units, pint.Quantity(200, units).to("A/V").magnitude)
    ...:
pA/V 2e-10
nA/V 2.0000000000000002e-07
uA/V 0.00019999999999999998
```

Note: this will make *pint* a new requirement for *apstools*

prjemian:
**If** the gain was an `ophyd.Signal` (or subclass), then it could be recorded in a run's baseline.

prjemian:
Setting a precedent for current amplifiers (pre-amps) here that `gain` is a `Component` (either RO or R/W) with name shortcut to the parent `Device`.  Per #544

PR #547, opened 2021-10-04
refactor from unittest to pytest

### LINKED ISSUE #447: switch unit testing to py.test
This will match what the bluesky framework is using

prjemian:
- FIX #447


prjemian:
Needs testing on APS workstation

prjemian:
Can anyone review by the end of this week?

prjemian:
This reorganizes the unit tests and begins testing with a real EPICS IOC.  It is needed for #549 so it would be most useful to merge before 2021-10-07.

apstools/beamtime/apsbss.py
```diff
@@ -167,7 +167,7 @@ def epicsUpdate(prefix):
     bss.status_msg.put("clearing PVs ...")
     bss.clear()
 
-    cycle = bss.esaf.aps_cycle.get()
+    cycle = bss.esaf.aps_cycle.get(timeout=10)
```
prjemian, line 170: Not a good idea to add an explicit timeout here in library code.  Better to set an overall timeout from the caller.

prjemian: [For example](https://github.com/BCDA-APS/bluesky_training/blob/main/bluesky/instrument/framework/initialize.py#L125-L128), this must be called before any ophyd instances are created:

```python
# set default timeout for all EpicsSignal connections & communications
EpicsSignalBase.set_defaults(
    auto_monitor=True, timeout=60, write_timeout=60, connection_timeout=60,
)
```

PR #549, opened 2021-10-05
PVPositioner for energy, theta, & wavelength in Kohzu Monochromator Device

### LINKED ISSUE #519: Use PVPositionerKohzu energy, wavelength, and theta
The [PVPositioner](https://blueskyproject.io/ophyd/reference/positioners.html#pvpositioner) class provides a `stop()` method.  This could be used for (virtual axes) energy, wavelength, and theta to support the KohzuSeqCtl code in EPICS.

Still, this does not detail an easy way to stop *only* the monochromator motors and not stop other motors in the same IOC.  Currently, the *best practice* is to stop all motors in the IOC using the `$(P)allstop` PV (where `$(P)` is the IOC's PV prefix).

_Originally posted by @prjemian in https://github.com/BCDA-APS/apstools/issues/508#issuecomment-868167134_

prjemian:
- FIX #519

### DIFF
--- a/apstools/_devices/kohzu_monochromator.py
+++ b/apstools/_devices/kohzu_monochromator.py
@@ -75,10 +75,10 @@ class KohzuSoftPositioner(PVPositioner):
 
     def _setup_move(self, position):
         """Move and do not wait until motion is complete (asynchronous)."""
-        self.log.debug('%s.setpoint = %s', self.name, position)
+        self.log.debug("%s.setpoint = %s", self.name, position)
         self.setpoint.put(position, wait=False)  # was: wait=True
         if self.actuate is not None:
-            self.log.debug('%s.actuate = %s', self.name, self.actuate_value)
+            self.log.debug("%s.actuate = %s", self.name, self.actuate_value)
             self.actuate.put(self.actuate_value, wait=False)
 
 
--- a/apstools/_devices/tests/test_kohzu_monochromator.py
+++ b/apstools/_devices/tests/test_kohzu_monochromator.py
@@ -33,11 +33,13 @@ class MyKohzu(KohzuSeqCtl_Monochromator):
             # all motors in range, no work to do, MUST yield something
             yield from bps.null()
             return
+        # fmt: off
         yield from bps.mv(
             self.m_theta, p_theta,
             self.m_y, p_y,
             self.m_z, p_z,
         )
+        # fmt: on
         yield from bps.sleep(1)  # allow IOC to react
         yield from bps.mv(self.operator_acknowledge, 1)
 


prjemian:
code that exercises this support:

```python
"""
file: /tmp/kohzu.py
"""

import logging
import time

from apstools.devices import KohzuSeqCtl_Monochromator
from bluesky import plan_stubs as bps
from bluesky import RunEngine
from ophyd import Component
from ophyd import EpicsMotor
from ophyd import EpicsSignal
from ophyd import Signal
from ophyd import PVPositioner

logger = logging.getLogger(__name__)


class MyKohzu(KohzuSeqCtl_Monochromator):
    m_theta = Component(EpicsMotor, "m45")
    m_y = Component(EpicsMotor, "m46")
    m_z = Component(EpicsMotor, "m47")

    def into_control_range(self, p_theta=2, p_y=-15, p_z=90):
        """
        Move the Kohzu motors into range so the energy controls will work.
        """
        yield from bps.mv(
            self.m_theta, p_theta,
            self.m_y, p_y,
            self.m_z, p_z,
        )
        yield from bps.sleep(1)  # allow IOC to react
        yield from bps.mv(self.operator_acknowledge, 1)

    def stop(self):
        self.m_theta.stop()
        self.m_y.stop()
        self.m_z.stop()


def notify(text):
    status_text.put(text)


status_text = EpicsSignal("gp:gp:text1", name="status_text")
dcm = MyKohzu("gp:", name="dcm")
dcm.wait_for_connection()
status_text.wait_for_connection()

notify("starting RunEngine ...")
RE = RunEngine({})
notify("moving into control range ...")
RE(dcm.into_control_range())

delay = 0.1
notify(f"sleep {delay} s")
time.sleep(delay)

notify(f"put into Auto mode & sleep {delay} s")
dcm.mode.put("Auto")
time.sleep(delay)

for target in (10.8, 10.2):
    notify(f"moving energy to {target:.3f} ...")
    dcm.energy.move(target)

notify(f"final energy: {dcm.energy.position:.3f} keV")
```

prjemian:
Note: gets stuck when sent to same positions where it is:

```python
%mov dcm.energy dcm.energy.setpoint.get()
```

Needs the *Move* button pressed in the GUI before it will complete the bluesky move command.

prjemian:
Still, this does not STOP properly from the RunEngine because it stuck is waiting on the setpoint's `put()` to be verified.  (in `_setup_move()`)

prjemian:
The Kohzu control screen uses the IOC's `allstop` PV to stop.  This will interrupt the move without having to know the actual motors involved.

PR #553, opened 2021-10-07
add unit tests for APS cycle signals

### LINKED ISSUE #552: unit test for datetime issue 550
The bug reported in #550 would have been found earlier if that code had been covered by a unit test.

The [ApsCycleComputedRO](https://github.com/BCDA-APS/apstools/blob/6a7163caade8a875d1407e8a6db7dec19c7d94e6/apstools/_devices/aps_cycle.py#L38) class can be tested without an EPICS IOC.

prjemian:
- FIX #552

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,16 @@ describe the future plans.
 
 :1.5.3:  release expected by 2021-11-01
 
+   * `#553 <https://github.com/BCDA-APS/apstools/pull/553>`_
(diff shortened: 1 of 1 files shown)


prjemian:
New tests pass on APS workstation.

@mdwyman Can you review this?  The test can be run (from root of repository using 2021-2 environment) with: `pytest -vvv apstools/_devices/tests`

PR #574, opened 2021-11-20
isolate dependency on aps-dm-api  package into apsbss

### LINKED ISSUE #570: Move the *apsbss* parts to a dedicated repository.
Move the *apsbss* parts to a dedicated repository.

The only strong dependencies for which the *apstools* package needs *aps-dm-api* are:

* the APS cycle
* hosts the *apsbss* package and application

The APS cycle is obtained from a table obtained through the APS data management API.  The table comes from the APS database server.  If (and only if) the `aps-dm-api` package is not available, could obtain the current cycle information by a local copy of that table.  Downside is that the local copy would need to be updated every run-cycle (or as frequently as the schedule table is changed).

_Originally posted by @prjemian in https://github.com/BCDA-APS/apstools/issues/561#issuecomment-974433237_

### LINKED ISSUE #571: Remove direct dependency on *aps-dm-api* package.  Get current APS cycle from call to apsbss or local table.
Remove direct dependency on *aps-dm-api* package to enable #570.  Get current APS cycle from call to apsbss or local table.

_Originally posted by @prjemian in https://github.com/BCDA-APS/apstools/issues/561#issuecomment-974475798_

prjemian:
- FIX #571


prjemian:
test failure is a time-zone issue

PR #581, opened 2021-11-26
remove apsbss

### LINKED ISSUE #570: Move the *apsbss* parts to a dedicated repository.

### LINKED ISSUE #582: workflow to build & publish Sphinx to GH pages
follow the lead of the new `apsbss` package here: https://github.com/BCDA-APS/apsbss/blob/main/.github/workflows/publish-sphinx.yml

prjemian:
- FIX #570
- FIX #582


prjemian:
#582 : The new workflow will publish a successful build of the docs to https://bcda-aps.github.io/apstools/ whenever a new tag is created.

Issue #618, opened 2022-01-12
add unit test for stop() method


prjemian:
https://github.com/BCDA-APS/apstools/blob/7645cf133bd6dadbd29a3a8f26f89c1fea4f4456/apstools/devices/positioner_soft_done.py#L197-L202

prjemian:
Re-opening since [related workflow still fails](https://github.com/BCDA-APS/apstools/runs/4845006875?check_suite_focus=true).

PR #624, opened 2022-01-17
Remove auto_heater and tweaks to PVPositionerSoftDone

gfabbris:
This is what I meant before about cleaning up the lakeshore support. This removes the `auto_heater` that is setup at 4-ID-D, which automatically changes the heater power according to the setpoint. I think this might be worth having, but probably best fit for the EPICS support.

I did some preliminary tests at 4-ID-D and it seems to be working fine. However, there is still the lingering issue of the LS340 setpoint not updating sometimes. Not sure if it happens only at sector 4.

I also did some minor tweaks to PVPositionerSoftDone that I realized might be needed while I was testing this.


gfabbris:
I should mention that one of the `PVPositionerSoftDone` tweaks was to set the startup value of `target` to -1. Bluesky seems to be unhappy if it is `None`.

apstools/devices/lakeshore_controllers.py
```diff
@@ -215,68 +160,9 @@ def __init__(self, *args, loop_number=None, **kwargs):
...
-    )
-
-    ramp_rate = FormattedComponent(
-        EpicsSignalWithRBV, "{prefix}RampR{loop_number}", kind="config"
-    )
-    ramp_on = FormattedComponent(
-        EpicsSignalWithRBV, "{prefix}OnRamp{loop_number}", kind="config"
-    )
+    @property
+    def egu(self):
```
prjemian, line 164: This is syntactic sugar but inconsistent in apstools.  Is it necessary to meet a particular interface?  Otherwise, it becomes a special case that needs additional explanation (with _this_ support, use `.egu` but with other devices, use `.units.get()`).

I'm requesting this property be removed unless absolutely necessary.

apstools/devices/lakeshore_controllers.py
```diff
@@ -302,11 +196,11 @@ class LakeShore336Device(Device):
...
 
-class LS340_LoopBase(PVPositionerSoftDoneWithStop):  # TODO: check, check, check
+class LS340_LoopBase(PVPositionerSoftDoneWithStop):
     """ Base settings for both sample and control loops. """
 
     # position
-    readback = Component(Signal, value=0)
+    # readback is provided by PVPositionerSoftDoneWithStop
     # TODO: Polar reports: sometimes the setpoint PV is not updated
     setpoint = FormattedComponent(
```
prjemian, line 205: setpoint is also provided by PVPositionerSoftDoneWithStop so this is confusing

apstools/devices/positioner_soft_done.py
```diff
@@ -144,7 +144,7 @@ def __init__(
         **kwargs,
     ):
 
-        if setpoint_pv == readback_pv:
+        if setpoint_pv == readback_pv and setpoint_pv != "":
```
prjemian, line 147: This additional logic causes a unit test to fail: https://github.com/BCDA-APS/apstools/blob/a3c3c92d8f6be4807024eb765ea69846d5774af6/apstools/devices/tests/test_positioner_soft_done.py#L45-L48

Is there a reason to allow both readback_pv and setpoint_pv to be empty strings?

apstools/devices/lakeshore_controllers.py
gfabbris: The issue here is that the write/read PVs are different, and that is not "allowed" in PVPositionerSoftDoneWithStop.

apstools/devices/positioner_soft_done.py
gfabbris: I came across this issue because I wanted to set `setpoint` and `readback` in the child class, so both `setpoint_pv` and `readback_pv` `kwargs` were `""`

prjemian: Aren't some additional checks necessary then so that PVPositionerSoftDone does not attempt to proceed with these empty strings AND an defined readback and/or setpoint Components?

See an example of what you describe in class [`apstools.devices.Eurotherm2216e`](https://github.com/BCDA-APS/apstools/blob/40a79e3ce6d70449e2979a193ed33380e1d504fa/apstools/devices/eurotherm_2216e.py#L32)

https://github.com/BCDA-APS/apstools/blob/40a79e3ce6d70449e2979a193ed33380e1d504fa/apstools/devices/eurotherm_2216e.py#L39-L46

and the constructor: https://github.com/BCDA-APS/apstools/blob/40a79e3ce6d70449e2979a193ed33380e1d504fa/apstools/devices/eurotherm_2216e.py#L79-L88

The unmodifed logic can handle this and prevent the error from empty strings with no Components.

prjemian: The [Eurotherm's unit test](https://github.com/BCDA-APS/apstools/blob/40a79e3ce6d70449e2979a193ed33380e1d504fa/apstools/devices/tests/test_eurotherm_2216e.py#L14) is trivial since the controller is not generally available to a unit test workflow.  https://github.com/BCDA-APS/apstools/blob/40a79e3ce6d70449e2979a193ed33380e1d504fa/apstools/devices/tests/test_eurotherm_2216e.py#L14-L31

However, it does test that no errors are generated when creating the object (including the setpoint and readback Components).  Connection with EPICS is not tested.

apstools/devices/lakeshore_controllers.py
prjemian: If you mean the two PVs have different prefixes, it is possible to use full PV names:

```python
pos = PVPositionerSoftDoneWithStop("", readback_pv="A:float1", setpoint_pv="b:parameter", name="pos"))
```

gfabbris: I meant for cases where there are different PVs for writing and reading a parameter. For example, for the setpoint,  I would like to use the "OUT1:SP_RBV" as the PV reading the value, and "OUT1:SP" as the write PV. I can do that in `EpicsSignal` with the `write_pv` `kwarg`.

Screen Shot 2022-01-19 at 11 37 23 AM

apstools/devices/positioner_soft_done.py
gfabbris: Perhaps we should then modify the `PVPositionerSoftDone` `__init__` to the one below?

```
    def __init__(
        self,
        prefix="",
        *,
        readback_pv="ignoreRBV",
        setpoint_pv="ignore",
        tolerance=None,
        update_target=True,
        **kwargs,
    ):
```

gfabbris: And modify the logic back to `if setpoint_pv == readback_pv:`

prjemian: That seems reasonable since a different signal class is desired.  Same decision as the Eurotherm controller support.

gfabbris:
One perhaps more general question about `PVPositionerSoftDone` is whether we should have the `setpoint_pv` and `readback_pv` `kwargs`, in other words, force this class to be aways inherited. This is how `PVPositioner` is defined in [ophyd](https://github.com/bluesky/ophyd/blob/df60483867e521dcda83648756cab8fc7b80dd17/ophyd/pv_positioner.py#L15).

gfabbris:
Not sure why the py3.7 failed...

gfabbris:
Voilà, it now passed...

prjemian:
> One perhaps more general question about `PVPositionerSoftDone` is whether we should have the `setpoint_pv` and `readback_pv` `kwargs`, in other words, force this class to be aways inherited. This is how `PVPositioner` is defined in [ophyd](https://github.com/bluesky/ophyd/blob/df60483867e521dcda83648756cab8fc7b80dd17/ophyd/pv_positioner.py#L15).

I like the ability provided by `PVPositioner` to override the `readback` and `setpoint` (and other) Components.  Swapping `EpicsSignalWithRBV` or `Signal` for `EpicsSignal`, for example.  We should do that here.  Yet, on the other hand... Do we have any obligation to existing use of `PVPositionerSoftDone` (in which case we should be reluctant to make change without strong reason)?  Yes, here in apstools, in support for temperature controllers and slits.

The only thing to change is the default value of `target`, change from `None` to `-1`.  If by `-1`, we mean _undefined_, there are other ways to do that (define an `UNDEFINED=object()` symbol and test for it).

prjemian:
> Not sure why the py3.7 failed... I'm re-running it just to check consistency

Guessing this was a random failure due to unit tests of the `stop()` method (#618).

apstools/devices/positioner_soft_done.py
```diff
@@ -137,8 +137,8 @@ def __init__(
         self,
         prefix="",
         *,
-        readback_pv="",
-        setpoint_pv="",
+        readback_pv="ignoreRBV",
```
prjemian, line 140: Not sure either of these changes is helpful, can they be removed?

gfabbris:
> > One perhaps more general question about `PVPositionerSoftDone` is whether we should have the `setpoint_pv` and `readback_pv` `kwargs`, in other words, force this class to be aways inherited. This is how `PVPositioner` is defined in [ophyd](https://github.com/bluesky/ophyd/blob/df60483867e521dcda83648756cab8fc7b80dd17/ophyd/pv_positioner.py#L15).
> 
> I like the ability provided by `PVPositioner` to override the `readback` and `setpoint` (and other) Components. Swapping `EpicsSignalWithRBV` or `Signal` for `EpicsSignal`, for example. We should do that here. Yet, on the other hand... Do we have any obligation to existing use of `PVPositionerSoftDone` (in which case we should be reluctant to make change without strong reason)? Yes, here in apstools, in support for temperature controllers and slits.
> 
> The only thing to change is the default value of `target`, change from `None` to `-1`. If by `-1`, we mean _undefined_, there are other ways to do that (define an `UNDEFINED=object()` symbol and test for it).

I'm not sure what you tried to argue here. Should we make it the same as `PVPositioner`? I think it does make the logic simpler.

prjemian:
I argued for both ways, concluding that changing `target=-1` is the only change we should make here.  (Other support expects this interface as-is.)

apstools/devices/positioner_soft_done.py
```diff
@@ -96,7 +97,7 @@ class PVPositionerSoftDone(PVPositioner):
     tolerance = Component(Signal, value=-1, kind="config")
     report_dmov_changes = Component(Signal, value=False, kind="omitted")
 
-    target = Component(Signal, value=-1, kind="config")
+    target = Component(Signal, value=UNDEFINED, kind="config")
```
prjemian, line 100: test for it!  in tests/test_positioner_soft_done.py

gfabbris: Still unhappy:

```python
ValueError: Cannot determine the appropriate bluesky-friendly data type for value
<object object at 0x7f15b845cd20> of Python type <class 'object'>.
Supported types include: int, float, str, and iterables such as list, tuple, np.ndarray, and so on.
```

gfabbris: what about np.nan?

prjemian: Aack!  Switch it back to `-1`.

gfabbris: What about "None" (string)?

prjemian: Ok.  It should be an internal value that the user won't see until they forget to define it.

gfabbris: This is only a problem until the positioner is moved for the first time

gfabbris:
FYI, just got this error in `def test_luascript_reset():`:

```
>       assert lua.inputs.BB.input_value.get() == ""
338
E       AssertionError: assert 'BB.input_value' == ''
339
E         + BB.input_value
340

341
/home/runner/work/apstools/apstools/apstools/synApps/tests/test_luascript.py:79: AssertionError
342
=============================== warnings summary ===============================
343
../../../../../usr/share/miniconda/envs/anaconda-test-env-py-3.8/lib/python3.8/site-packages/pims/image_reader.py:26
344
  /usr/share/miniconda/envs/anaconda-test-env-py-3.8/lib/python3.8/site-packages/pims/image_reader.py:26: RuntimeWarning: PIMS image_reader.py could not find scikit-image. Falling back to matplotlib's imread(), which uses floats instead of integers. This may break your scripts. 
345
  (To ignore this warning, include the line "warnings.simplefilter("ignore", RuntimeWarning)" in your script.)
346
    warnings.warn(RuntimeWarning(ski_preferred))
347

348
-- Docs: https://docs.pytest.org/en/stable/warnings.html
349
=========================== short test summary info ============================
350
FAILED apstools/synApps/tests/test_luascript.py::test_luascript_reset - Asser...
351
============= 1 failed, 264 passed, 1 warning in 117.77s (0:01:57) =============
```

prjemian:
Transient error that disturbs me.  Why do we need `time.sleep(SHORT_DELAY_FOR_EPICS)`?  Seems to be a delay on our end to wait for database processing in the EPICS IOC.  Isn't `put_complete=True` supposed to fix that?

prjemian (approved):
LGTM

PR #681, opened 2022-07-29
add advice about 'blocking_callbacks' in Pilatus example

### LINKED ISSUE #671: blocking_callbacks in area detector setup
@MarkRivers noticed in `docs/source/examples/_ad__pilatus.rst`:

```py
OrderedDict([('enable', 1),
            ('blocking_callbacks', 'Yes'),
```

> It is really not good practice to be setting blocking_callbacks to Yes.  That slows down the detector.  The preferred practice is to set blocking_callbacks=No in the plugin and set wait_for_plugins=Yes in the driver.  That way all plugins can run in their own threads without slowing down the driver, and the detector will still wait for them all to complete before it reports that acquisition is done.

Thanks!  I'll label this as a _bug_ here.

prjemian:
- close #671
- refs #696


prjemian:
CI failure is unrelated (related to #680).

prjemian:
@MarkRivers - If you have time, I'd appreciate your review of these changes.  Thanks for the contribution.

MarkRivers:
I'm not sure enough of the code is shown here to know if it will work.  What does Ophyd do to wait for the detector to be done?  Ideally it uses ca_put_callback, which in pyepics is pv.put(wait=true).  The callback will not occur until the detector and all plugins are complete if WaitForPlugins=1.  It is not sufficient to wait for ADAcquire to go to 0 because that will happen as soon as the detector itself it done.  If the code is polling then it can wait for AcquireBusy to go to 0.  That is not as robust as ca_put_callback because you need to make sure that you first wait for AcquireBusy to go to 1, and then back to 0, so you are sure acquisition has started before deciding that it is done. That is difficult with very short acquire times, and leads to ad-hoc delays being added.

MarkRivers (commented):
I'm not sure enough of the code is shown here to know if it will work. What does Ophyd do to wait for the detector to be done? Ideally it uses ca_put_callback, which in pyepics is pv.put(wait=true). The callback will not occur until the detector and all plugins are complete if WaitForPlugins=1. It is not sufficient to wait for ADAcquire to go to 0 because that will happen as soon as the detector itself it done. If the code is polling then it can wait for AcquireBusy to go to 0. That is not as robust as ca_put_callback because you need to make sure that you first wait for AcquireBusy to go to 1, and then back to 0, so you are sure acquisition has started before deciding that it is done. That is difficult with very short acquire times, and leads to ad-hoc delays being added.

prjemian:
@MarkRivers : Thanks for the comments.  I agree that offering some proof that this works would be a good addition.  I will convert this static documentation to a Jupyter notebook.  Local development and execution should result in demonstration of an example proven to work.

prjemian:
The requested changes are not related specifically to the Pilatus detector.  The intent of this example was to show the support specific for the Pilatus all in one document.  Also include writing HDF5 files.

My access to a Pilatus is restricted now and certainly not able to access one as a Jupyter notebook documentation would require.  I will create an example with the ADSimDetector that writes an HDF5 file, showing the goals of this document.  Then, this document may be refactored to show the modifications necessary for the Pilatus.  The similar document for the [Perkin-Elmer](https://bcda-aps.github.io/apstools/examples/_ad_pe.html) should be reviewed at that time.

prjemian:
From the notebook developed so far, it is obvious that acquisition finished before all the frames were written to the HDF5 file (3 written out of 5 acquired).

Make a replacement for `ophyd.areadetector.SingleTrigger` called `SingleTriggerAcquireBusy` to implement the suggestion.

prjemian:
@MarkRivers - Following your suggestions, I have revised the [first example](https://github.com/BCDA-APS/apstools/blob/671-blocking-callbacks-in-AD-setup/docs/source/examples/_ad_adsim_hdf5_basic.ipynb), including updates to the underlying support to add `WaitForPlugins` and `AcquireBusy` in the driver (`cam` here) and `PLUGIN.blocking_callbacks = "No"`.  The example notebook is now based on ADSimDetector and uses the standard ophyd uid-based file naming convention (such as: `/tmp/example/2022/08/10/6b3a199a-e08a-44b5-ad79_000000.h5`).  Another example will show how to let EPICS control the file name.  That is in line with our beam lines expectations.

I would appreciate if you have time to comment on this new documentation.  Following that, I'll make the new document with the EPICS-controlled file names, then revise the examples for the Pilatus and Perkin-Elmer cameras.

prjemian:
One CI failed repeatedly, here:

```py
        assert not adsimdet.cam.is_busy
    
>       assert plugin.full_file_name.get().strip() not in ("", "/")
E       AssertionError: assert '/' not in ('', '/')
E        +  where '/' = <built-in method strip of str object at 0x7fe5378fbeb0>()
E        +    where <built-in method strip of str object at 0x7fe5378fbeb0> = '/'.strip
E        +      where '/' = <bound method EpicsSignalBase.get of EpicsSignalRO(read_pv='ad:TIFF1:FullFileName_RBV', name='adsimdet_tiff1_full_file_name', parent='adsimdet_tiff1', value='/', timestamp=1660149768.595167, auto_monitor=True, string=True)>()
E        +        where <bound method EpicsSignalBase.get of EpicsSignalRO(read_pv='ad:TIFF1:FullFileName_RBV', name='adsimdet_tiff1_full_file_name', parent='adsimdet_tiff1', value='/', timestamp=1660149768.595167, auto_monitor=True, string=True)> = EpicsSignalRO(read_pv='ad:TIFF1:FullFileName_RBV', name='adsimdet_tiff1_full_file_name', parent='adsimdet_tiff1', value='/', timestamp=1660149768.595167, auto_monitor=True, string=True).get
E        +          where EpicsSignalRO(read_pv='ad:TIFF1:FullFileName_RBV', name='adsimdet_tiff1_full_file_name', parent='adsimdet_tiff1', value='/', timestamp=1660149768.595167, auto_monitor=True, string=True) = AD_EpicsFileNameTIFFPlugin(prefix='ad:TIFF1:', name='adsimdet_tiff1', parent='adsimdet', read_attrs=[], configuration_attrs=[]).full_file_name

/home/runner/work/apstools/apstools/apstools/devices/tests/test_area_detector_support.py:204: AssertionError
=========================== short test summary info ============================
FAILED apstools/devices/tests/test_area_detector_support.py::test_bp_count_custom_name[tiff1-AD_EpicsFileNameTIFFPlugin]
============ 1 failed, 385 passed, 54 skipped in 186.23s (0:03:06) =============
```

This happens occasionally, that one of the file writing plugins (HDF5, JPEG, or TIFF) does not provide a full file name.  Why?  What makes this fail?  What makes this fail _randomly_?

prjemian:
Simply re-run the test in CI.  It passed on the third run.

prjemian:
@MarkRivers - CI is now passing with some consistency.  The documents have been revised with your guidance on `blacking_callbacks = "No"`, `wait_for_plugins = "Yes"`, and use of `acquire_busy` to indicate end of acquisition.

I'm ready to merge this unless you have further comments.

MarkRivers:
That looks OK.  Where is the polling for acquire_busy actually done?

prjemian:
The polling loop is in a thread (Python's threading package) and waiting (time.sleep) until the CA monitor reports the new value (0).  The separate thread is so the wait does not block the RunEngine from its other tasks.  The RunEngine is told to wait for this thread to finish (but handle its other tasks) before continuing.

prjemian:
Let me adjust that comment since it applies to other situations such as waiting for a custom signal.  

In this case (`trigger()` of an area detector's `cam` component: `detector.cam.trigger()`), the [default ophyd `trigger` code](https://github.com/bluesky/ophyd/blob/2fa3d6d271e928a3fe217845c3b0be5d4eb0523f/ophyd/areadetector/trigger_mixins.py#L136-L147):

```py
    def trigger(self):
        "Trigger one acquisition."
        if self._staged != Staged.yes:
            raise RuntimeError(
                "This detector is not ready to trigger."
                "Call the stage() method before triggering."
            )

        self._status = self._status_type(self)
        self._acquisition_signal.put(1, wait=False)
        self.generate_datum(self._image_name, ttime.time(), {})
        return self._status
```

The [`self._status`](https://github.com/bluesky/ophyd/blob/master/ophyd/status.py) is an object that manages the thread for this trigger.

The default ophyd algorithm is that the PV being triggered (`:cam1:Acquire`) is the same PV being waited on in the thread.  A [callback routine](https://github.com/bluesky/ophyd/blob/2fa3d6d271e928a3fe217845c3b0be5d4eb0523f/ophyd/areadetector/trigger_mixins.py#L149-L156) is previously configured (during `detector.cam.stage()`) to receive CA monitors on the `self._acquisition_signal`:

```py
    def _acquire_changed(self, value=None, old_value=None, **kwargs):
        "This is called when the 'acquire' signal changes."
        if self._status is None:
            return
        if (old_value == 1) and (value == 0):
            # Negative-going edge means an acquisition just finished.
            self._status.set_finished()
            self._status = None
```

So I changed the `stage()` method in the apstools package here to use `:cam1:AcquireBusy` as the PV to wait (and _not_ the `self._acquisition_signal`): https://github.com/BCDA-APS/apstools/blob/2884ed451f6ca2b7461ec10fb4d998c624af8efd/apstools/devices/area_detector_support.py#L754-L766

MarkRivers:
OK, I understand.  That seems like the right way to do it.

PR #728, opened 2022-10-05
Fix PVPositionerSoftDone stuck problem

### LINKED ISSUE #725: PVPositionerSoftDone gets stuck if target position is the same as current position
I ran into this problem with a couple of devices that use `PVPositionerSoftDone`. I think this is because if the `position` is the same as the current value in:

https://github.com/BCDA-APS/apstools/blob/c52de8902827955160e56ff25a9345559422784a/apstools/devices/positioner_soft_done.py#L191

Then `cb_setpoint` is not called, and `done` is never switched to `False`.

prjemian:
- ref #725


prjemian:
CI is not running, why?

prjemian:
No [CI](https://github.com/BCDA-APS/apstools/actions/workflows/unit_tests.yml)?  Why?

apstools/devices/tests/test_positioner_soft_done.py
```diff
@@ -362,3 +364,30 @@ def test_move_calcpos(target, calcpos):
...
+    # Confirm the initial position is as expected.
+    if calcpos.position == target:
+        # First, move away from the target.
+        status = calcpos.move(-target or 0.1 * (1 + random.random()))
+        assert status.elapsed > 0, str(status)
+        confirm_in_position(calcpos)
+
+    # Move to the target position.
+    status = calcpos.move(target)
+    assert status.elapsed > 0, str(status)
```
gfabbris, line 387: I'm not sure what the `assert str(status)` tests. Do we need to check for `status.done == True`?

prjemian:
Ok, call off the dogs, the [GHA status](https://www.githubstatus.com/) is _degraded_ at this time.  There is a _Partial Outage_.  Time for lunch.

prjemian:
I'll merge the unit tests but keep #725 open, pending more information about how to reproduce the problem.

PR #730, opened 2022-10-07
ignore Lakeshore336 setpoint's RBV

### LINKED ISSUE #725: PVPositionerSoftDone gets stuck if target position is the same as current position

prjemian:
- close #725


apstools/devices/positioner_soft_done.py
```diff
@@ -191,6 +191,9 @@ def _setup_move(self, position):
         if self.actuate is not None:
             self.log.debug("%s.actuate = %s", self.name, self.actuate_value)
             self.actuate.put(self.actuate_value, wait=False)
+        # This is needed because in a special case the setpoint.put does not
+        # run the "sub_value" subscriptions.
+        self.cb_setpoint()
```
prjemian, line 196: Will this callback be called twice in some cases?

prjemian: The `cb_setpoint()` method can inspect the `*args, **kwargs` to learn how it was called.

prjemian: These two lines could be run twice: https://github.com/BCDA-APS/apstools/blob/e91fa01a6797766d4a72c367917ef45a3fb3f4e3/apstools/devices/positioner_soft_done.py#L146-L147

The first looks like that would have no consequence, the second could stop a move.

gfabbris: I think it will not stop the motion. It could false-flag the motion as not done, but the `self.cb_readback()` in the following line forces a position check.

prjemian:
@gfabbris Is this ready to merge?

prjemian:
Getting a local CI fail in unrelated code:

```py
        assert calcs.connected
        if round(motor.position, 2) != 0:
            motor.move(0)
        result = listdevice(motor, scope=scope)
        assert isinstance(result, pd.DataFrame)
        assert column in result
        assert row in result[column]
>       assert result[column][row] == value
E       assert 5.0 == 1.0

apstools/utils/tests/test_listdevice.py:130: AssertionError
=============================================================================== short test summary info ===============================================================================
FAILED apstools/utils/tests/test_listdevice.py::test_spotchecks[epics-6-value-1.0] - assert 5.0 == 1.0
========================================================================== 1 failed, 309 deselected in 4.32s ==========================================================================
```

Might be related to the state of local IOC.  Nevertheless, good to resolve before merge.

prjemian:
For sure it is local to the IOC now.  It fails the test for motor velocity set to 1.0.  Currently, due to other work (#731), the velocity is 5.0.  Test some other floating-point motor parameter.

prjemian:
@gfabbris  I plan merge this tomorrow, to make a bug fix release (including this and #732).

apstools/devices/lakeshore_controllers.py
```diff
@@ -44,7 +44,7 @@ class LakeShore336_LoopControl(PVPositionerSoftDoneWithStop):
     # readback is provided by PVPositionerSoftDoneWithStop
 
     setpoint = FormattedComponent(
-        EpicsSignalWithRBV, "{prefix}OUT{loop_number}:SP", put_complete=True
+        EpicsSignal, "{prefix}OUT{loop_number}:SP", put_complete=True
```
gfabbris, line 47: I think we should reverse this, it's better to have the RBV, and it should work with the fix in the `PVPositionerSoftDone`.

prjemian: I'll make the change and test it.

prjemian: passes locally

prjemian:
Merged documentation changes (in `CHANGES.rst`) from #732.

prjemian:
@gfabbris Ready for review.

gfabbris (approved):

prjemian:
@gfabbris Thanks!

PR #733, opened 2022-10-12
revise packaging for included/excluded content

### LINKED ISSUE #732: Are yml files missing from packaging?
User sends by email:

> I was on 1.6.3 because when I installed 1.6.5 from conda-forge it complained about missing some yml files. Installing from source does not have the same issue.

Similar report about pip.

Investigate.

prjemian:
- close #732


prjemian:
Even after merge, will need to check package built by conda-forge.  Use a rc-versioned process to test.

PR #786, opened 2023-01-27
NXWriter updates

### LINKED ISSUE #781: NeXus callback and area detectors: can't create link to detector data
Message received over slack.com, Nikea channel:

Hugo Valim wrote:

> I'm using the nexus callback and it works fine with simple detectors. But when I use area detectors in scans it doesn't work because it can't create the link to the detector data. Basically it can not open the file to make the link. I'm thinking this is more a misconfiguration in the detector abstraction in OPHYD than a problem with the callback itself, is that right? Have you seen this before? Thank you very much!

### LINKED ISSUE #785: NXWriter callback: two problems
Note a couple problems with the output file (observed while working on #781).

* [x] The image (`adsimdet_image/value`) has a wrong attribute value: `@source_address = "/entry/data/data"`.  Should be `"/entry/data/adsimdet_image"`.
* [x] The `/entry/instrument/detectors` group is empty.

_Originally posted by @prjemian in https://github.com/BCDA-APS/apstools/issues/781#issuecomment-1406744739_

prjemian:
- close #785

### DIFF
--- a/apstools/callbacks/nexus_writer.py
+++ b/apstools/callbacks/nexus_writer.py
@@ -93,9 +93,8 @@ class NXWriter(FileWriterCallbackBase):
         decide if a signal in the primary stream is a detector or a positioner
         """
         try:
-            primary = self.root[
-                "/entry/instrument/bluesky/streams/primary"
-            ]
+            k = "/entry/instrument/bluesky/streams/primary"
+            primary = self.root[k]
         except KeyError:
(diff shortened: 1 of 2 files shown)


prjemian:
New `linter` workflow generates [many](https://github.com/BCDA-APS/apstools/actions/runs/4028184615/jobs/6924786248) `isort` and `black` code-style reports.

PR #795, opened 2023-02-01
NXWriter must wait until external data stream file is readable

### LINKED ISSUE #781: NeXus callback and area detectors: can't create link to detector data

prjemian:
- close #781

@HugoValim: Thanks for reporting this!  Feel free to comment on the PR.


prjemian:
CI is failing when `camera` is staged.

But, more important, the idea here has some a consequence.  When NXWriter delays the `writer()` method to a thread, this means a new acquisition cannot proceed until the thread finishes, such as: https://github.com/BCDA-APS/apstools/blob/e159bbb0bc24fb5f56862e83f84b0f26940eab21/apstools/callbacks/tests/test_nxwriter.py#L121-L122

With this idea in place, any user must implement a wait (such as this while loop) for the callback to finish before proceeding with analysis or a next acquisition.

Need to explore a different way to resolve the deadlock:

* area detector IOC keeps HDF5 file open until ophyd device is unstaged
* NXWriter callback runs to completion before run ends
* ophyd devices are unstaged just after the run ends

prjemian:
@HugoValim - I must seek some advice before proceeding with work on issue #781.

prjemian:
CI runs successfully with no errors now 4x in sequence.  Random CI failures resolved.

Issue #905, opened 2024-01-08
Occasional pytest fails in CI, yet CI reported as OK


prjemian:
CI failed in code not related to this PR.  Coincident with such failure is dramatic reduction on code coverage.  Here is the Py3.11 log with the fail

```py
2024-01-08T21:23:21.2348097Z apstools/plans/tests/test_alignment.py::test_lineup2[signals4-mover4--1.2-1.2-11-max-2] FAILED [ 48%]
2024-01-08T21:23:21.2349107Z 
2024-01-08T21:23:21.2349402Z =================================== FAILURES ===================================
2024-01-08T21:23:21.2350418Z _______________ test_lineup2[signals4-mover4--1.2-1.2-11-max-2] ________________
2024-01-08T21:23:21.2352695Z 
2024-01-08T21:23:21.2354570Z signals = [SynPseudoVoigt(name='pvoigt', value=5, timestamp=1704748988.992658), EpicsSignalRO(read_pv='gp:userCalc1.VAL', name='..., 'channels.chan06.gate', 'count_mode', 'delay', 'auto_count_delay', 'freq', 'preset_time', 'auto_count_time', 'egu'])]
2024-01-08T21:23:21.2358614Z mover = EpicsSignal(read_pv='gp:gp:float1', name='axis', value=0.0, timestamp=1704748999.251151, tolerance=0.0001, auto_monitor=True, string=False, write_pv='gp:gp:float1', limits=False, put_complete=False)
2024-01-08T21:23:21.2360720Z start = -1.2, finish = 1.2, npts = 11, feature = 'max', nscans = 2
2024-01-08T21:23:21.2368328Z 
2024-01-08T21:23:21.2372382Z     @pytest.mark.parametrize(
2024-01-08T21:23:21.2377464Z         "signals, mover, start, finish, npts, feature, nscans",
2024-01-08T21:23:21.2378157Z         [
2024-01-08T21:23:21.2378841Z             [noisy, m1, -1.2, 1.2, 11, "max", 2],  # slower, is motor
2024-01-08T21:23:21.2379855Z             [pvoigt, axis, -1.2, 1.2, 11, "cen", 2],  # faster, is ao (float)
2024-01-08T21:23:21.2380802Z             [pvoigt, axis, -1.2, 1.2, 11, "max", 2],
2024-01-08T21:23:21.2381613Z             [[pvoigt], axis, -1.2, 1.2, 11, "max", 2],  # list
[...]

PR #913, opened 2024-01-22
rename the workflow files: simple, generic names

prjemian:


prjemian:
Workflow failures will be resolved in separate usse (#905).  This PR only renames the workflow files.

PR #917, opened 2024-01-25
SR-570 race condition

canismarko:
Fixes a race condition where the SR570 pre-amp gets the sensitivity_value set before the sensitivity_unit, ending up at the wrong gain.

If the gain is set to 1 mA/V in the IOC, and you try to set it to e.g. 200 µA/V, you need to set the unit first then the value since "1" is the only valid "mA/V" value:

✓ 1 mA/V -> 1 µA/V -> 200 µA/V
✗ 1 mA/V -> 200 mA/V (corrects to 1 mA/V) -> 1 µA/V

Solved this problem by using ``put_complete=True`` for the *sensitivity_unit* signal.

### DIFF
--- a/apstools/devices/srs570_preamplifier.py
+++ b/apstools/devices/srs570_preamplifier.py
@@ -182,7 +182,7 @@ class SRS570_PreAmplifier(PreamplifierBaseDevice):
     # in the EPICS .db file.  Must cast them to ``float()`` or ``int()``
     # as desired.
     # see: https://github.com/epics-modules/ip/blob/master/ipApp/Db/SR570.db
-    sensitivity_value = Component(GainSignal, "sens_num", kind="config", string=True, put_complete=True)
+    sensitivity_value = Component(GainSignal, "sens_num", kind="config", string=True)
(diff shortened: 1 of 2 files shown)


canismarko:
Want to check this one last time at the beamline, then it will be ready for review.

prjemian:
Tests ran with errors here.

canismarko:
> Tests ran with errors here.

You mean locally? The CI looks like it's just that weird decrease in test coverage.

prjemian:
> Tests ran with errors here.

Meant to say: "Tests ran with **_no new_** errors here."

canismarko:
Finally had a chance to test this at the beamline. Totally didn't forget about it and just noticed it was still in draft while I was looking for something else. Anyway, I think it works, best as I can tell.

I swapped the gain back and forth a hundred or so times in a while loop, and they always got set in the right order.

canismarko:
Any thoughts about merging this PR?

prjemian (approved):
LGTM

prjemian:
@canismarko - You can merge now

PR #918, opened 2024-02-09
DOC: minor edits to documentation about exception handling

tacaswell:


prjemian:
@tacaswell Thanks for posting this.

docs/source/examples/ho_contingency.ipynb
```diff
@@ -296,13 +301,22 @@
...
     "\n",
     "decorator | synopsis\n",
     "--- | ---\n",
     "`finalize_decorator` | Simple. Runs the `final_plan` no matter what happens in the decorated plan.\n",
     "`contingency_decorator` | Full-featured. Handle each aspect of Python's `try..except..else..finally` clause.\n",
     "\n",
+    "**Exception handling in generator coroutines**\n",
+    "\n",
+    "Generator coroutines [expose exception throwing](https://tacaswell.github.io/coroutines-i.html) as a user communication channel.  This means\n",
+    "there is some [in-band encoding](https://youtu.be/iKzOBWOHGFE?si=XAtQKhk3eHboHcL-&t=1011) of Python's control exceptions with the users exceptions.  When the `close()` [method](https://docs.python.org/3/reference/expressions.html#generator.close) is called on generator corutine Python will throw a  [`GeneratorExit`](https://docs.python.org/3/library/exceptions.html#GeneratorExit) exception into the coroutine.  If you catch this exception and try to yield another `Msg`, either explictily in an `except` block or in a `finally` block Python will raise a `RuntimeError` at the call site.\n",
```
prjemian, line 315: small typos

```suggestion
    "there is some [in-band encoding](https://youtu.be/iKzOBWOHGFE?si=XAtQKhk3eHboHcL-&t=1011) of Python's control exceptions with the users exceptions.  When the `close()` [method](https://docs.python.org/3/reference/expressions.html#generator.close) is called on a generator coroutine, Python will throw a  [`GeneratorExit`](https://docs.python.org/3/library/exceptions.html#GeneratorExit) exception into the coroutine.  If you catch this exception and try to yield another `Msg`, either explicitly in an `except` block or in a `finally` block, Python will raise a `RuntimeError` at the call site.\n",
```

docs/source/examples/ho_contingency.ipynb
```diff
@@ -296,13 +301,22 @@
...
+    "\n",
+    "If you want to directly use `try..except..else..finally` you must handle this case.  See the source of these decorators for a guide.\n",
```
prjemian, line 317:
```suggestion
    "If you want to use `try..except..else..finally` directly, you must handle this case.  See the source of these decorators for a guide.\n",
```

prjemian (approved):
Thanks!

PR #920, opened 2024-02-15
lineup2() should work with low intensity peaks

### LINKED ISSUE #919: lineup2: problem for peaks of height 1
This code: https://github.com/BCDA-APS/apstools/blob/883f82c4914a5bdc777f130817befb96329230c1/apstools/plans/alignment.py#L332

should divide by some measure of the**Y** axis, not X

Thanks Victoria Cooley at APS  for the report!

prjemian:
- close #919

### DIFF
--- a/apstools/plans/alignment.py
+++ b/apstools/plans/alignment.py
@@ -333,8 +333,8 @@ def lineup2(
             return abs(value) > peak_factor
         except ZeroDivisionError:  # not enough samples
             try:
-                value = abs(stats.max_y / stats.min_y)
-                return value > peak_factor
+                value = stats.max_y / stats.min_y
+                return abs(value) > peak_factor
             except ZeroDivisionError:
                 return False
 
--- a/apstools/plans/alignment.py
+++ b/apstools/plans/alignment.py
@@ -329,8 +329,8 @@ def lineup2(
     def strong_peak(stats) -> bool:
         """Determine if the peak is strong."""
         try:
-            value = (stats.max_y - stats.min_y) / stats.sigma
-            return value > peak_factor
+            value = (stats.max_y - stats.min_y) / stats.stddev_y
+            return abs(value) > peak_factor
         except ZeroDivisionError:  # not enough samples
             try:
                 value = abs(stats.max_y / stats.min_y)
--- a/apstools/plans/tests/test_alignment.py
+++ b/apstools/plans/tests/test_alignment.py
@@ -100,10 +100,7 @@ def test_SynPseudoVoigt_randomize():
     assert signal.noise_multiplier == 1
 
 
-TestParameters = collections.namedtuple(
-    "TestParameters",
-    "signal mover start finish npts feature"
-)
+TestParameters = collections.namedtuple("TestParameters", "signal mover start finish npts feature")
 parms_slower = TestParameters(noisy, m1, -1.2, 1.2, 11, "max")
 parms_faster = TestParameters(pvoigt, axis, -1.2, 1.2, 11, "max")
 parms_cen = TestParameters(pvoigt, axis, -1.2, 1.2, 11, "cen")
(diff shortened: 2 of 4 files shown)


apstools/plans/alignment.py
```diff
@@ -329,8 +329,8 @@ def principal_signal_stats() -> str:
     def strong_peak(stats) -> bool:
         """Determine if the peak is strong."""
         try:
-            value = (stats.max_y - stats.min_y) / stats.sigma
-            return value > peak_factor
+            value = (stats.max_y - stats.min_y) / stats.stddev_y
+            return abs(value) > peak_factor
         except ZeroDivisionError:  # not enough samples
             try:
                 value = abs(stats.max_y / stats.min_y)
```
MDecarabas, line 336: Do we want to keep the format uniform across tries. Line 322 we return abs, line 336 we do the abs before return. Doesn't really matter in the grand scheme but I think it would be nice

apstools/plans/tests/test_alignment.py
```diff
@@ -285,3 +289,54 @@ class ImitatorForTesting(Device):
...
+        [-1e5, -1e4, 1e-8, 0.1, 0.2, -0.7, 0.5, 11, 1, 0.1, True],
+        [1e5, -10, 10, 0.1, 0.2, -0.7, 0.5, 11, 1, 0.05, True],
+        [1e-5, 0, 1e-8, 0.1, 0.2, -0.7, 0.5, 11, 1, 0.05, True],
+        [1e-5, 1e-7, 1e-8, 0.1, 0.2, -0.7, 0.5, 11, 1, 0.05, True],
+        [-1e5, 0, 1e-8, 0.1, 0.2, -0.7, 0.5, 11, 1, 0.05, True],
+        [0, 0, 1e-8, 0.1, 0.2, -0.7, 0.5, 11, 1, 0.05, False],  # only noise
+        [0, 1, 0.1, 0.1, 0.2, -0.7, 0.5, 11, 1, 0.05, False],  # bkg + noise
+        [0, 1, 100, 0.1, 0.2, -0.7, 0.5, 11, 1, 0.05, False],  # bkg + big noise
+        [0, 0, 0, 0.1, 0.2, -0.7, 0.5, 11, 1, 0.005, None],  # no signal  ZeroDivisionError
+        [0, 1, 0, 0.1, 0.2, -0.7, 0.5, 11, 1, 0.005, None],  # bkg only  ZeroDivisionError
```
MDecarabas, line 310: I don't know if we ever reuse these values however I think it might be useful/cleaner to store each list in a named variable as opposed to having such a list of lists. Hard to look at and understand what we are seeing, especially if we will be modifying test values. 

Ex:
```
signal_high_background = [1e5, 1e6, 10, 0.1, 0.2, -0.7, 0.5, 11, 1, 0.05, False]
```

MDecarabas (changes requested):
Small comments, mostly related to style

apstools/plans/alignment.py
prjemian: Good point.

apstools/plans/tests/test_alignment.py
prjemian: If these are defined as individual lists before the test, it seems easy to dissociate the values with the name.  Could a namedtuple be used?  If so, the fields of the namedtuple could be the first argument, then a list of the namedtuples would be the second.

apstools/plans/alignment.py
prjemian: See commit 58d7230e

MDecarabas (approved):
Smart choice using the named tuples. All looks good to me now!

apstools/plans/tests/test_alignment.py
prjemian: One more commit coming ...

prjemian: I like this approach (251bc1ea).  It forces the person writing the test parameters to explain, in some way, why that set of parameters is used.  The namedtuple and the type annotation is the Pythonic way to enforce that the same names are used for all parametrized tests.

prjemian: We can make that SOP going forward, and backfill the old tests by best efforts.

PR #958, opened 2024-04-08
mesh_grid feature for 8-ID

MDecarabas:


apstools/plans/xpcs_mesh.py
```diff
@@ -0,0 +1,237 @@
...
+from bluesky import utils
+from bluesky import plan_patterns
+from bluesky import preprocessors as bpp
+
+import inspect
+from itertools import zip_longest
+from collections import defaultdict
+import os
+
+try:
```
prjemian, line 11: Neither `cytools` or `toolz` is listed as a project requirement.  Should not assume that other packages will bring them in.  Add one or the other to this section (sorted alphabetically): https://github.com/BCDA-APS/apstools/blob/b0868619ba9450670903e26438af4dabf117c1af/pyproject.toml#L54-L72

apstools/plans/xpcs_mesh.py
```diff
@@ -0,0 +1,237 @@
...
+    # cytools is a drop-in replacement for toolz, implemented in Cython
+    from cytools import partition
+except ImportError:
+    from toolz import partition
+
+
+def mesh_list_grid_scan(detectors, *args, number_of_collection_points, snake_axes=False, per_step=None, md=None):
+    """
+    Scan over a mesh; each motor is on an independent trajectory.
```
prjemian, line 20: Sounds like a grid scan.  Needs a bit more.  Suggestion:

```
Scan over a multi-dimensional mesh, collecting a total of *n* points.

- Each motor is on an independent trajectory.
- Repeat the sequence of mesh positions until *n* points are collected.
```

prjemian: Is the mesh truly multi-dimensional or just 2-D?  Whatever, we should say.

apstools/plans/xpcs_mesh.py
```diff
@@ -0,0 +1,237 @@
...
+    :func:`bluesky.plans.list_scan`
+    :func:`bluesky.plans.rel_list_scan`
+    """
+
+    full_cycler = plan_patterns.outer_list_product(args, snake_axes)
+
+    md_args = []
+    motor_names = []
+    motors = []
+    for i, (motor, pos_list) in enumerate(partition(2, args)):
```
prjemian, line 65: If not using `i` in this loop, then no need for `enumerate()`.

apstools/plans/xpcs_mesh.py
```diff
@@ -0,0 +1,237 @@
...
+        "plan_pattern_args": dict(args=md_args, snake_axes=snake_axes),
+        "plan_pattern_module": plan_patterns.__name__,
+        "motors": tuple(motor_names),
+        "hints": {},
+    }
+    _md.update(md or {})
+    try:
+        _md["hints"].setdefault("dimensions", [(m.hints["fields"], "primary") for m in motors])
+    except (AttributeError, KeyError):
+        ...
```
prjemian, line 86: It would be useful/informative to call `logger.warning()` for these exceptions.  Capture the exception reason and report it to the logger:  `except (AttributeError, KeyError) as reason:`

apstools/plans/xpcs_mesh.py
```diff
@@ -0,0 +1,237 @@
...
+        "plan_args": {"detectors": list(map(repr, detectors)), "cycler": repr(cycler), "per_step": repr(per_step)},
+        "plan_name": "scan_nd",
+        "hints": {},
+    }
+    _md.update(md or {})
+    try:
+        dimensions = [(motor.hints["fields"], "primary") for motor in cycler.keys]
+    except (AttributeError, KeyError):
+        # Not all motors provide a 'fields' hint, so we have to skip it.
+        pass
```
prjemian, line 137: logger here, too

prjemian: maybe?

apstools/plans/xpcs_mesh.py
```diff
@@ -0,0 +1,237 @@
...
+        elif _verify_nd_step(sig):
+            # check other signature for back-compatibility
+            pass
+        elif _verify_1d_step(sig):
+            # Accept this signature for back-compat reasons (because
+            # inner_product_scan was renamed scan).
+            dims = len(list(cycler.keys))
+            if dims != 1:
+                raise TypeError(
+                    "Signature of per_step assumes 1D trajectory " "but {} motors are specified.".format(dims)
```
prjemian, line 198:
```suggest
                    f"Signature of per_step assumes 1D trajectory but {dims} motors are specified."
```

apstools/plans/xpcs_mesh.py
```diff
@@ -0,0 +1,237 @@
...
+        _md["hints"].setdefault("dimensions", dimensions)
+
+    predeclare = per_step is None and os.environ.get("BLUESKY_PREDECLARE", False)
+    if per_step is None:
+        per_step = bps.one_nd_step
+    else:
+        # Ensure that the user-defined per-step has the expected signature.
+        sig = inspect.signature(per_step)
+
+        def _verify_1d_step(sig):
```
prjemian, line 153: _verify_1d_step() and _verify_nd_step() are almost the same.  Can they be combined if you pass in the list for `zip_longest(the_list, items)`?

apstools/plans/xpcs_mesh.py
```diff
@@ -0,0 +1,237 @@
...
+    @bpp.stage_decorator(list(detectors) + motors)
+    @bpp.run_decorator(md=_md)
+    def scan_until_completion():
+        """
+        Scanning until the total number of required collection points is achieved
+        """
+        if predeclare:
+            yield from bps.declare_stream(*motors, *detectors, name="primary")
+        iterations = 0
+        while iterations < number_of_collection_points:
```
prjemian, line 230: I believe the outer `while` loop may be eliminated using `itertools.cycle` on top of the `cycler()`.  See this SO post: https://stackoverflow.com/a/23416519/1046449

Like this:

```py
        iterations = 0
        for step in itertools.cycle(cycler):
            yield from per_step(detectors, step, pos_cache)
            iterations += 1
            if iterations >- number_of_collection_points:
                break
```

prjemian (commented):
Couple things to examine.  On second thought, the logger.warning() for missing hints might be unwelcome.

apstools/plans/xpcs_mesh.py
MDecarabas: added cytools to pyproject.toml

apstools/plans/xpcs_mesh.py
MDecarabas: As copied from bluesy list_grid_scan, can change it but that is the way it is currently standardized. Please advise.

apstools/plans/xpcs_mesh.py
MDecarabas: As copied from bluesy list_grid_scan, can change it but that is the way it is currently standardized. Please advise.

apstools/plans/xpcs_mesh.py
MDecarabas: As copied from bluesy list_grid_scan, can change it but that is the way it is currently standardized. Please advise.

apstools/plans/xpcs_mesh.py
MDecarabas: As copied from bluesy list_grid_scan, can change it but that is the way it is currently standardized. Please advise.

apstools/plans/xpcs_mesh.py
MDecarabas: As copied from bluesy list_grid_scan, can change it but that is the way it is currently standardized. Please advise.

apstools/plans/xpcs_mesh.py
MDecarabas: Given, that it is copied from list_grid_scan I believe it is truly multi-dimensional. Not sure how it would look in higher n-dimensions but theoretically it should work.

apstools/plans/xpcs_mesh.py
prjemian: Let's change it.  The line was originally written for older Python.  Now all the supported versions use f-strings and there is no need for this to look so awkward.  Also the two strings ("" "") were originally on separate lines but the black formatter has put them on one line.  Time to merge those two strings.

apstools/plans/xpcs_mesh.py
prjemian: Can they be combined?  I'm suggesting that our opinion can supersede what the legacy support has done.  It means less code to review/revise/debug.

prjemian:
Can you finish this today?

prjemian:
One more addition: https://github.com/BCDA-APS/apstools/blob/ccecf2f65f0e3a2badb6ab9b3c848f9bac518ab7/environment.yml#L36

After this line, add:

```yml
 - toolz
````

prjemian (approved):
Great!  Assume the pin on toolz is needed.  Could synchronize same pinning in both `pyproject.toml `and `environment.yml`.  Not a blocker, though.

PR #987, opened 2024-06-28
Make it easier to use area detectors with a factory

### LINKED ISSUE #984: area detector factory function
A factory function might make it easier to create area detector classes and instances.  The factory would address common configurations (such as PVA, HDF5, ...) as options, triggered by keywords.  Additional kwargs would describe supplemental options such as the file path seen by the IOC.

prjemian:
- close #984


prjemian:
The initial commit is for a fixed version of ophyd area detector support.  It is templated on the support provided for the APS XPCS instrument.

prjemian:
Still need to add/blend [contribution](https://github.com/BCDA-APS/apstools/issues/984#issuecomment-2192536006) by @cooleyv that requests the ADCore version from the existing PV and [use case](https://github.com/BCDA-APS/apstools/issues/984#issuecomment-2195201893) from @canismarko

The challenge with asking the ADCore version is that the IOC must be running enough to respond.

prjemian:
Will be a fun challenge blending these and then building unit tests.

apstools/devices/area_detector_support.py
```diff
@@ -825,6 +826,186 @@ def unstage(self):
...
+    * ``generate_datum()`` - coordinate image storage metadata
+    """
+
+    def stage(self):
+        self.stage_sigs.move_to_end("capture", last=True)
+        super().stage()
+
+
+def ad_class_factory(
+    cam_class,
```
prjemian, line 860: Change to kwarg, default to CamBase

prjemian:
The factory method looks too complex.  Can it be simpler?  ([advice](https://www.geeksforgeeks.org/class-factories-a-powerful-pattern-in-python/))

prjemian:
There are very few detectors in use at APS that still use ADCore releases older than 3.4.  Additional support for versions of the various plugins is not justified.  They can be addressed individually if the new factory design is not adequate.

prjemian:
Needs some tests for [`instantiate_fake_device()`](https://github.com/BCDA-APS/apstools/issues/984#issuecomment-2195201893), as suggested by @canismarko.

prjemian:
Instead of: https://github.com/BCDA-APS/apstools/blob/abd00fe59fa63170b2dc6e6b3161a4499bbe1827/apstools/devices/area_detector_factory.py#L147-L148
to remove a default, such as: https://github.com/BCDA-APS/apstools/blob/abd00fe59fa63170b2dc6e6b3161a4499bbe1827/apstools/devices/area_detector_factory.py#L53-L60

allow the caller to re-define a key with `defaults` kwarg.  This kwargs will update the `PLUGIN_DEFAULTS` dictionary: https://github.com/BCDA-APS/apstools/blob/abd00fe59fa63170b2dc6e6b3161a4499bbe1827/apstools/devices/area_detector_factory.py#L82

prjemian:
@cooleyv, @keenanlang, @canismarko, @gfabbris, @strempfer, @MarkRivers  -- The new area detector factory support is ready for review.

prjemian:
@sureshnaps : Your suggestion has led to simplifications such as.

```py
    from ophyd.areadetector import SimDetectorCam
    from apstools.devices import ad_creator

    det = ad_creator(
        "ad:", name="det", class_name="MySimDetector",
        [
            {"cam": {"class": SimDetectorCam}},
            "image",  # all the defaults are acceptable
            "pva",
            {"hdf1": {"write_path_template": "/", "read_path_template": "/mnt/ioc/tmp"}},
            "codec1",
            "proc1",
            "roi1",
            "stat1",
        ],
    )
```

prjemian:
In the `PLUGIN_DEFAULTS`, the origin (ophyd _v_. apstools) of the default class for each plugin should be more obvious.  Use absolute paths for the classes from ophyd.

MDecarabas (approved):
Looks good, can you make an issue to remove the local version out of 8-ID and transition to the APS tools version. We can make the changes at the same time we re-implement specwriter

apstools/devices/tests/test_issue984_ad_factory.py
canismarko, line 1: This pretty much captures how I use this feature. I'll keep an eye out for edge cases where this doesn't work, but I can't think of any right now.

canismarko (approved):
This strikes me as a positive change.

The difference between the *plugins* and *plugin_defaults* arguments to the AD factory is tricky for me to understand. I think this is because of the complicated way area detectors are built in ophyd rather than a design issue for the factory, so probably this needs to be well documented.

apstools/devices/tests/test_issue984_ad_factory.py
prjemian: I auditioned a few names for `plugin_defaults` before deciding on a name which matched the `PLUGIN_DEFAULTS` in the module.  The general idea here, naming aside, is that the default plugin classes decided by the module from apstools could be replaced completely by the caller.  One advantage is future-proofing, as newer versions of the support classes appear.  Another case that came up in testing directly was how to remove the `write_path_template` key from `hdf1`.  It seems more natural to be able to edit or replace the `plugin_defaults` dictionary.

Documentation should clarify this.

Might better names for `plugins` and `plugin_defaults` help to distinguish their purpose?  I'm thinking `config` instead of `plugins`, for example.

canismarko: I don't have better suggestions for the naming. I think the name *plugins* does accurately reflect what the argument is doing, since in its simplest use it really is just a list of plugins to include. I think keeping *plugin_defaults* to match ``PLUGIN_DEFAULTS`` makes sense too.

One easy change might be to use keyword arguments in the example at the top of `area_detector_factory.py` so it's clear what is *plugins* and what is *plugin_defaults*, at least for the example that includes both. On first read I got lost in the curly and square brackets, making it harder to understand.

apstools/devices/area_detector_factory.py
```diff
@@ -0,0 +1,330 @@
...
+   ~ad_creator
+   ~ad_class_factory
+   ~PLUGIN_DEFAULTS
+
+EXAMPLES
+
+Just the camera plugin (uses CamBase, the most basic features)::
+
+    from apstools.devices import ad_creator
+    det = ad_creator("ad:", name="det", class_name="MySimpleAD", ["cam",])
```
canismarko, line 16: *plugins* needs to be a keyword argument for this to work.

apstools/devices/area_detector_factory.py
```diff
@@ -0,0 +1,330 @@
...
+
+View ADSimDetector image with CA and PVA::
+
+    from ophyd.areadetector import SimDetectorCam
+    from apstools.devices import ad_creator
+
+    det = ad_creator(
+        "ad:", name="det", class_name="MySimDetector",
+        [
```
canismarko, line 25: Make a kw argument

apstools/devices/area_detector_factory.py
```diff
@@ -0,0 +1,330 @@
...
+
+Record HDF5 images with Eiger detector. Here, both the Eiger detector IOC and
+the Bluesky databroker use the same filesystem mount ``/``::
+
+    from ophyd.areadetector import EigerDetectorCam
+    from apstools.devices import ad_creator
+
+    det = ad_creator(
+        "ad:", name="det", class_name"MyEiger",
+        [
```
canismarko, line 40: Make a kw argument.

apstools/devices/area_detector_factory.py
```diff
@@ -0,0 +1,330 @@
...
+
+    from ophyd.areadetector import EigerDetectorCam
+    from apstools.devices import ad_creator, PLUGIN_DEFAULTS
+
+    plugin_defaults = PLUGIN_DEFAULTS.copy()
+    plugin_defaults["hdf1"].pop("read_path_template", None)
+    plugin_defaults["hdf1"].pop("write_path_template", None)
+
+    det = ad_creator(
+        "ad:", name="det", class_name"MyEiger",
```
canismarko, line 59: Make into kw arguments.

canismarko (changes requested):
Updating my reivew based on Pete's point that these actually need to be keyword arguments in the examples.

apstools/devices/tests/test_issue984_ad_factory.py
canismarko: Ah yes, I think you would get a "positional argument follows keyword argument" error as-is. I updated my review.

apstools/devices/area_detector_factory.py
prjemian: This is the most problematic of the `plugin_defaults` list structure.  It won't work as written since `plugin_defaults` is a Python `list`.  Which means a unit test should be provided.

canismarko: ``PLUGIN_DEFAULTS`` is a dict (https://github.com/BCDA-APS/apstools/blob/61c93bbc81a67b46d61855a57433a6756c20831e/apstools/devices/area_detector_factory.py#L82), and *plugin_defaults* is type-hinted and used as a dict.

prjemian: Right.  Exactly _that_ kind of confusion.

prjemian: Example is correct as written, once the `plugins=` is added.

apstools/devices/area_detector_factory.py
```diff
@@ -0,0 +1,330 @@
...
+        *object*:
+        Plugin configuration dictionary.
+        (default: ``None``, PLUGIN_DEFAULTS will be used.)
+    """
+    bases = bases or (
+        SingleTrigger_V34,
+        ophyd.areadetector.DetectorBase,
+    )
+    plugins = plugins or ["cam"]
+    plugin_defaults = plugin_defaults or PLUGIN_DEFAULTS
```
canismarko, line 244: This makes it impossible to pass an empty dict (``{}``) to this function as *plugin_defaults*, since ``bool({})`` evaluates to ``False`` and then ``PLUGIN_DEFAULTS`` would get used.

Maybe not a big deal since I can't imagine many people would want this. But I think it could be avoided by using ``PLUGIN_DEFAULTS`` as the default value for the *plugin_defaults*, which makes the intention more clear when looking at the call signature.

prjemian: Can do.

prjemian: Same for `plugins=None`.  Changing it to `plugins=["cam"]`.

prjemian: Ah... there's the reason to keep the default kwarg value of `None` -- in the unit tests, `None` is used.

```py
ERROR apstools/devices/tests/test_area_detector_support.py::test_AD_EpicsFileNameMixin[hdf1-AD_HDF5] - TypeError: Must be a dict.  Received plugin_defaults=None
ERROR apstools/devices/tests/test_area_detector_support.py::test_AD_EpicsFileNameMixin[jpeg1-AD_JPEG] - TypeError: Must be a dict.  Received plugin_defaults=None
ERROR apstools/devices/tests/test_area_detector_support.py::test_AD_EpicsFileNameMixin[tiff1-AD_TIFF] - TypeError: Must be a dict.  Received plugin_defaults=None
ERROR apstools/devices/tests/test_area_detector_support.py::test_stage[hdf1-AD_EpicsFileNameHDF5Plugin] - TypeError: Must be a dict.  Received plugin_defaults=None
ERROR apstools/devices/tests/test_area_detector_support.py::test_stage[jpeg1-AD_EpicsFileNameJPEGPlugin] - TypeError: Must be a dict.  Received plugin_defaults=None
ERROR apstools/devices/tests/test_area_detector_support.py::test_stage[tiff1-AD_EpicsFileNameTIFFPlugin] - TypeError: Must be a dict.  Received plugin_defaults=None
ERROR apstools/devices/tests/test_area_detector_support.py::test_bp_count_custom_name[hdf1-AD_EpicsFileNameHDF5Plugin] - TypeError: Must be a dict.  Received plugin_defaults=None
ERROR apstools/devices/tests/test_area_detector_support.py::test_bp_count_custom_name[jpeg1-AD_EpicsFileNameJPEGPlugin] - TypeError: Must be a dict.  Received plugin_defaults=None
ERROR apstools/devices/tests/test_area_detector_support.py::test_bp_count_custom_name[tiff1-AD_EpicsFileNameTIFFPlugin] - TypeError: Must be a dict.  Received plugin_defaults=None
ERROR apstools/devices/tests/test_area_detector_support.py::test_full_file_name_local - TypeError: Must be a dict.  Received plugin_defaults=None
ERROR apstools/devices/tests/test_area_detector_support.py::test_no_file_path - TypeError: Must be a dict.  Received plugin_defaults=None
ERROR apstools/devices/tests/test_area_detector_support.py::test_file_numbering - TypeError: Must be a dict.  Received plugin_defaults=None
ERROR apstools/devices/tests/test_area_detector_support.py::test_capture_error_with_single_mode - TypeError: Must be a dict.  Received plugin_defaults=None
ERROR apstools/devices/tests/test_area_detector_support.py::test_single_mode - TypeError: type.__new__() argument 2 must be tuple, not None
```

Code like this is needed:

```py
    if plugin_defaults is None:
        plugin_defaults = PLUGIN_DEFAULTS
```

canismarko: Okay, sounds good.

prjemian: > I can't imagine many people would want `plugin_defaults={}`

Actually, this is a realistic case to force the caller to define all keywords of each plugin.  Probably only useful for unit testing, not real world?

prjemian: @canismarko Will these changes work for you?

canismarko (approved):

apstools/devices/area_detector_factory.py
canismarko: Yes, changes look good to me. Thanks!

prjemian: I also added some documentation (look for `EXAMPLE 1:` and others).

PR #988, opened 2024-06-29
Wrote new support for the planar undulator.

canismarko:
The new support uses positioners, so the following sub-devices can be moved directly.

- energy
- gap
- energy_taper
- gap_taper

I removed the ApsDualUndulator class for now because it's not clear how to specify the kind of undulators for each branch, and it would be trivial for beamline's to create this device themselves. I'm not opposed to adding it back if there 's a good solution, though.

I tested this on the downstream ID at 25-ID.

### DIFF
--- a/apstools/devices/aps_undulator.py
+++ b/apstools/devices/aps_undulator.py
@@ -8,21 +8,20 @@ APS undulator
    ~ApsUndulatorDual
 """
 
+import logging
+from enum import IntEnum
+
 from ophyd import Component
+from ophyd import DerivedSignal
 from ophyd import Device
 from ophyd import EpicsSignal
 from ophyd import EpicsSignalRO
+from ophyd import PVPositioner
 from ophyd import Signal
 
 from .tracking_signal import TrackingSignal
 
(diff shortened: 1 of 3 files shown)


apstools/devices/aps_undulator.py
```diff
@@ -8,81 +8,93 @@
...
-    .. index:: Ophyd Device; ApsUndulator
+class DoneStatus(IntEnum):
+    MOVING = 0
+    DONE = 1
 
-    EXAMPLE::
 
-        undulator = ApsUndulator("ID09ds:", name="undulator")
-    """
+class BusyStatus(IntEnum):
```
prjemian, line 32: BusyStatus belongs with the synApps busy record support.

prjemian: https://github.com/BCDA-APS/apstools/blob/main/apstools/synApps/busy.py

prjemian (approved):
LGTM

prjemian:
@canismarko -- Thanks for the contribution!

Issue #970, opened 2024-04-29
aps_cycle.py


prjemian:
- https://github.com/BCDA-APS/apstools/blob/ece96632cb919c34c18bc200645189a5632fa576/apstools/devices/aps_cycle.py#L22
- https://github.com/BCDA-APS/apstools/blob/ece96632cb919c34c18bc200645189a5632fa576/apstools/devices/aps_cycle.py#L117

prjemian:
Verify that this code works in the APS-U era.

prjemian:
Might need the new PV gateway as environment variable.

> [Tuesday 10:52 AM] Lang, Keenan C.
> That's what I did for /net/s8iddserv/xorApps/epics/ui/startup/start_epics  to have it show the new current
 > And from my work computer
> `klang@kobold:~$ export EPICS_CA_AUTO_ADDR_LIST=no`
> `klang@kobold:~$ export EPICS_CA_ADDR_LIST="pvgatemain1.aps4.anl.gov"`
> `klang@kobold:~$ caget S-DCCT:CurrentM`
>` S-DCCT:CurrentM                -0.00471643`

MDecarabas:
I believe the issue can be closed. This is the output when testing. Looks as expected. Let me know if there is any other specific testing you want me to run


![Image](https://github.com/BCDA-APS/apstools/assets/72196105/84096b7f-98fb-4411-9274-a95b79ade5cd)

prjemian:
Great!  Now I see the method was actually processed in the info report.

PR #994, opened 2024-07-01
update APS operating cycle info

### LINKED ISSUE #971: aps_cycle_info.yml
Last entry now: https://github.com/BCDA-APS/apstools/blob/ece96632cb919c34c18bc200645189a5632fa576/apstools/devices/aps_cycle_info.yml#L137-L139

prjemian:
- close #971 
- #970 (necessary revisions were not caught in this PR)

### DIFF
--- a/apstools/devices/aps_cycle.py
+++ b/apstools/devices/aps_cycle.py
@@ -9,12 +9,14 @@ APS cycles
 
 import datetime
 import json
+import logging
 import pathlib
 import time
 
 import yaml
 from ophyd.sim import SynSignalRO
 
+logger = logging.getLogger(__name__)
 _PATH = pathlib.Path(__file__).parent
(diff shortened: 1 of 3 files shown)


apstools/devices/aps_cycle.py
```diff
@@ -103,11 +108,23 @@ def sorter(line):
 
     @property
     def _bss_list_runs(self):
+        """Get the full list of APS runs via a Data Management API or 'None'."""
+        from dm import ApsDbApiFactory
+        from dm.common.exceptions.dmException import DmException
+        from ..utils.aps_data_management import dm_source_environ
```
MDecarabas, line 114: does dm not get used elsewhere?

MDecarabas (approved):

apstools/devices/aps_cycle.py
prjemian: Yes.  But what is the concern here?

PR #1002, opened 2024-07-08
2M & 4M undulators

### LINKED ISSUE #975: 2M insertion device (only 01DS)
image

### LINKED ISSUE #976: 4M insertion device (only 11US and 11DS)
image

prjemian:
- close #975
- close #976

NOTE: 2M & 4M -- No PV for Done Moving, rely on Busy and set `done_value=0` instead.

@cooleyv
@cpchuang

### DIFF
--- a/apstools/devices/aps_undulator.py
+++ b/apstools/devices/aps_undulator.py
@@ -8,6 +8,7 @@ APS undulators (Insertion Devices)
    ~Revolver_Undulator
    ~STI_Undulator
    ~Undulator2M
+   ~Undulator4M
 """
 
 import logging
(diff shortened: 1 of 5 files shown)


prjemian:
Random and unrelated EPICS CA connection timeout problems in testing.

prjemian:
On 3rd (or 4th?) re-run of all tests, they passed.  Might adjust master timeout if failures continue:
https://github.com/BCDA-APS/apstools/blob/ded5219b092ecd6cc8f112a209b76baa8fa66945/apstools/tests/__init__.py#L12

Applied here:
https://github.com/BCDA-APS/apstools/blob/ded5219b092ecd6cc8f112a209b76baa8fa66945/apstools/tests/__init__.py#L27-L32

prjemian:
Testing is more stable with the longer master timeout.  3s was borderline to encounter EPICS CA connection timeouts.

prjemian:
@MDecarabas Now, this is ready for review.

apstools/devices/aps_undulator.py
```diff
@@ -178,6 +183,54 @@ class STI_Undulator(PlanarUndulator):
...
+
+    EXAMPLE::
+
+        undulator = Undulator4M("S11ID:DSID:", name="undulator")
+    """
+
+    # PVs not found
+    busy = None
+    magnet = None
+    version_plc = None
```
MDecarabas, line 227: Isn't this the same content with just a different class name?

MDecarabas (commented):
Small question about if we can just reuse the 2m undulator class for the 4m as well. Might have missed something but it does not look like any of the code is different

prjemian:
Thought about this, too.  Similar to the `STI_Undulator`?  Once (or if) we identify differences, then we can break the dependency.

MDecarabas:
> Thought about this, too. Similar to the `STI_Undulator`? Once (or if) we identify differences, then we can break the dependency.

Are you saying as the class will grow you expect to see differences?

prjemian:
I'm saying that as the class is used at the beam lines, we might discover signals needed by one but not the others.

apstools/devices/aps_undulator.py
prjemian: Made 4M as a subclass of 2M

MDecarabas (approved):

PR #1013, opened 2024-08-26
resolve RuntimeError from listdevice() with an area detector object

### LINKED ISSUE #990: RuntimeError from listdevice()
image

prjemian:
- close #990


prjemian:
Root cause: area detector IOC has just started and no image data has been passed from `cam` to `image` plugins.  Image shape tuple is all zeroes.  This is raised by `ophyd.areadetector.base`.  We can trap that so that `listdevice()` will not fail.  User will still see a `Subscription value callback exception` that is internal to `ophyd.areadetector.base`.  

Nothing we can do about that here.  User should be advised to _acquire an image_ after which the error message will not appear.

MDecarabas (commented):
Looks good, let's add somewhere a place about how to use this function. Maybe inside the doctoring

MDecarabas (approved):

prjemian:
The workflow failure after the merge will be resolved by another PR.

- #1016

PR #1005, opened 2024-07-22
Updates to PVPostionerSoftDone

gfabbris:
Main proposed changes:

- Keeps `setpoint` and `readback` as normal `EpicsSignal`. But now, changes to the `setpoint` are sent to the `target` through a [subscription](https://github.com/gfabbris/apstools/blob/4ef2ab5d92d915cbd620dca836d246d89383796e/apstools/devices/positioner_soft_done.py#L141).
- [Forces](https://github.com/gfabbris/apstools/blob/4ef2ab5d92d915cbd620dca836d246d89383796e/apstools/devices/positioner_soft_done.py#L229) the `done` signal to be `False` when the device is moved.
- Changes the default value of `use_target` to `False`. There is one disadvantage of using the `target` signal: if you change the setpoint "manually" in EPICS (and the device moves to the new setpoint), the Bluesky device will not know that it is `inposition` because `target` was not updated. Therefore, I recommend to set use_target=True only when EPICS changes the setpoint value while ramping (like in the Lakeshore 340).
- Removes python 3.8 from unit tests. It is leading to an import error, see [error](https://github.com/gfabbris/apstools/actions/runs/10044498274/job/27760058119#step:15:94).

PR #1008, opened 2024-08-17
Write data to SPEC files point-by-point

prjemian:
- close #1006

PR #1014, opened 2024-08-27
Revert "Updates to PVPostionerSoftDone"

gfabbris:
Reverts BCDA-APS/apstools#1005

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -99,7 +99,7 @@ jobs:
     strategy:
       matrix:
         python-version:
-          # - "3.8"  # TODO: Breaking
+          - "3.8"
           - "3.9"
           - "3.10"
           - "3.11"
--- a/apstools/devices/eurotherm_2216e.py
+++ b/apstools/devices/eurotherm_2216e.py
@@ -83,7 +83,7 @@ class Eurotherm2216e(PVPositionerSoftDoneWithStop):
             readback_pv="ignoreRBV",
             setpoint_pv="ignore",
             tolerance=tolerance,
-            use_target=False,
+            update_target=False,
             **kwargs,
         )
         self.sensor.subscribe(self.cb_sensor)
--- a/apstools/devices/lakeshore_controllers.py
+++ b/apstools/devices/lakeshore_controllers.py
@@ -68,7 +68,7 @@ class LakeShore336_LoopControl(PVPositionerSoftDoneWithStop):
 
     def __init__(self, *args, loop_number=None, timeout=10 * HOUR, **kwargs):
         self.loop_number = loop_number
-        super().__init__(*args, timeout=timeout, tolerance=0.1, use_target=True, readback_pv=f"IN{loop_number}", **kwargs)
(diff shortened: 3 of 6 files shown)


gfabbris:
@prjemian: I think this will work. You can actually see the changes here, and we can just close this if we want to keep it.

prjemian:
> we can just close this if we want to keep it.

not sure I understand.  Can you rephrase?

prjemian:
This branch did not trigger CI yet.  It passed `pytest -vvv ./apstools/devices` locally with no failures.

gfabbris:
I mean that, since it was already merged into the 916-PVPositionerSoftDone-update branch, if you want to keep these changes, then we don't merge this PR.

gfabbris:
By the way, forgot the add the comments I had for this:

Main proposed changes:

- Keeps `setpoint` and `readback` as normal `EpicsSignal`. But now, changes to the `setpoint` are sent to the `target` through a [subscription](https://github.com/gfabbris/apstools/blob/4ef2ab5d92d915cbd620dca836d246d89383796e/apstools/devices/positioner_soft_done.py#L141).
- [Forces](https://github.com/gfabbris/apstools/blob/4ef2ab5d92d915cbd620dca836d246d89383796e/apstools/devices/positioner_soft_done.py#L229) the `done` signal to be `False` when the device is moved.
- Changes the default value of `use_target` to `False`. There is one disadvantage of using the `target` signal: if you change the setpoint "manually" in EPICS (and the device moves to the new setpoint), the Bluesky device will not know that it is `inposition` because `target` was not updated. Therefore, I recommend to set use_target=True only when EPICS changes the setpoint value while ramping (like in the Lakeshore 340).
- Removes python 3.8 from unit tests. It is leading to an import error, see [error](https://github.com/gfabbris/apstools/actions/runs/10044498274/job/27760058119#step:15:94).

gfabbris:
> This branch did not trigger CI yet. It passed `pytest -vvv ./apstools/devices` locally with no failures.

I think these are the CI, got triggered when I accidentally merged: https://github.com/BCDA-APS/apstools/actions/runs/10568053789

gfabbris:
It looks like this [error](https://github.com/gfabbris/apstools/actions/runs/10044498274/job/27760058119#step:15:94) was due to a bug in `setuptools` that got [fixed](https://github.com/pypa/setuptools/pull/4486), it works now (https://github.com/gfabbris/apstools/actions/runs/10584246206/job/29328167045?pr=4). Should we add it back?

prjemian:
> It looks like this [error](https://github.com/gfabbris/apstools/actions/runs/10044498274/job/27760058119#step:15:94) was due to a bug in `setuptools` that got [fixed](https://github.com/pypa/setuptools/pull/4486), it works now (https://github.com/gfabbris/apstools/actions/runs/10584246206/job/29328167045?pr=4). Should we add it back?

It was not a problem when the SpecWriterCallback2 was developed:

- https://github.com/BCDA-APS/apstools/pull/1008

prjemian:
@gfabbris : With #954 merged, can you close this PR?

Issue #1022, opened 2024-10-10
PVPositionerSoftDone uses invalid subscription event type for simulated devices


canismarko:
In apstools v 1.7.0, a new subscription was added that uses the *event_type="setpoint"*. In a real device, this works fine, but for simulated devices this fails because "setpoint" is not a valid event type.

```python
In [1]: from apstools.devices import PVPositionerSoftDone

In [2]: from ophyd.sim import instantiate_fake_device

In [3]: instantiate_fake_device(PVPositionerSoftDone, readback_pv="spam:", setpoint_pv="eggs:")
---------------------------------------------------------------------------
UnknownSubscription                       Traceback (most recent call last)
Cell In[3], line 1
----> 1 instantiate_fake_device(PVPositionerSoftDone, readback_pv="sldkjf", setpoint_pv="lldkjf")

File /home/beams1/B268176/miniforge3/envs/haven-test2/lib/python3.10/site-packages/ophyd/sim.py:1307, in instantiate_fake_device(dev_cls, name, prefix, **specified_kw)
   1305 kwargs["name"] = name if name is not None else dev_cls.__name__
   1306 kwargs["prefix"] = prefix
-> 1307 return dev_cls(**kwargs)

File /home/beams1/B268176/miniforge3/envs/haven-test2/lib/python3.10/site-packages/apstools/devices/positioner_soft_done.py:141, in PVPositionerSoftDone.__init__(self, prefix, readback_pv, setpoint_pv, tolerance, use_target, **kwargs)
    139 self.readback.subscribe(self.cb_readback)
    140 self.setpoint.subscribe(self.cb_setpoint)
--> 141 self.setpoint.subscribe(self.cb_update_target, event_type="setpoint")
    143 # cancel subscriptions before object is garbage collected
    144 weakref.finalize(self.readback, self.readback.unsubscribe_all)

File /home/beams1/B268176/miniforge3/envs/haven-test2/lib/python3.10/site-packages/ophyd/ophydobj.py:481, in OphydObject.subscribe(self, callback, event_type, run)
    479 # check that this is a valid event type
    480 if event_type not in self.subscriptions:
--> 481     raise UnknownSubscription(
    482         "Unknown subscription {!r}, must be one of {!r}".format(
    483             event_type, self.subscriptions
    484         )
    485     )
    487 # wrapper for callback to snarf exceptions
    488 def wrap_cb(cb):

[...]

PR #1027, opened 2024-10-23
allow package import even if databroker v2.0+ is installed in environment

### LINKED ISSUE #1026: certain imports block work with databroker v2
Certain imports, such as https://github.com/BCDA-APS/apstools/blob/6692d5b1c9e0d20223c7b985ece539d0522ec58b/apstools/utils/list_runs.py#L23-L25
block any new or test work with databroker v2.  These imports are not used later, it seems.

These definitions could be relocated, closer to the code in which they are used: https://github.com/BCDA-APS/apstools/blob/6692d5b1c9e0d20223c7b985ece539d0522ec58b/apstools/utils/_core.py#L12-L22

@canismarko , Thanks for alerting about this.

prjemian:
- close #1026

Changes should be transparent to current operations but should allow development and testing (such as with [`ophyd-async`](https://github.com/bluesky/ophyd-async)) if databroker is installed.

### DIFF
--- a/apstools/devices/aps_data_management.py
+++ b/apstools/devices/aps_data_management.py
@@ -35,8 +35,6 @@ from ophyd import Component
 from ophyd import Device
 from ophyd import Signal
 
-from ..utils import run_in_thread
-
 logger = logging.getLogger(__name__)
 
 NOT_AVAILABLE = "-n/a-"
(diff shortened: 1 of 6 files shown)


prjemian:
@canismarko @MDecarabas  Ready for review.

canismarko (approved):

canismarko:
Changes look good. I tested Haven against this new version, only additional exceptions are coming from #1022, which should go away once I convert our slits over to ophyd-async.

Thanks!

PR #1063, opened 2025-03-27
WIP: Feature/Add_typehints and guarnieri stubs

ravescovi:


apstools/__init__.py
```diff
@@ -2,15 +2,17 @@
 
 """Configuration of apstools package."""
 
+from typing import Optional
```
prjemian, line 5: Where will `Optional` be used in this module?

apstools/callbacks/callback_base.py
```diff
@@ -10,8 +10,9 @@
 
 import datetime
 import logging
-import pathlib
+from typing import Any, Dict, List, Optional, Union
```
prjemian, line 13: Is there an advantage to use `typing.Dict[str, str]` when `dict[str, str]` passes linting and testing?  Same for `typing.List`.

apstools/callbacks/callback_base.py
```diff
@@ -60,16 +61,16 @@ class FileWriterCallbackBase:
        ~stop
     """
 
-    file_extension = "dat"
-    _file_name = None
-    _file_path = None
+    file_extension: str = "dat"
+    _file_name: Optional[pathlib.Path] = None
```
prjemian, line 65: For both of these, is `Optional[Union[pathlib.Path,str]]` acceptable?  I think so.  Check the unit testing to see if both ways are tested.

ravescovi: Updated. Checking the tests soon.

apstools/callbacks/callback_base.py
ravescovi: I don't think there is any effect difference here. Here is why I started using typing as default:
https://stackoverflow.com/questions/37087457/difference-between-defining-typing-dict-and-dict

apstools/callbacks/callback_base.py
ravescovi: Created a new test. Need to check if is redundant with others.

prjemian:
With #1067 resolved, merge `main` into this branch and see if the CI passes.

apstools/callbacks/callback_base.py
ravescovi: Forget my previous post. After 3.9 or 3.10 this are the same. I will move back to standard types.

canismarko:
Maybe you realized this already, but guarneri already has types. Shouldn't need stubs.

prjemian:
```py
2025-04-25T16:56:29.1964943Z E   ImportError: cannot import name 'JJtransfocator1x' from 'apstools.devices.jj_transfocators' (/home/runner/work/apstools/apstools/apstools/devices/jj_transfocators.py)
```

Need to merge main branch into this one to get the `jj_transfocators.py` code.  Code is missing from this branch:
image

ravescovi:
Leave this PR here while I break this into smaller branches.

PR #1088, opened 2025-05-14
a device that just combines a bunch of motors using just kwargs

### LINKED ISSUE #1086: Device for a bunch of motors, defined using just kwargs
In https://github.com/APS-4ID-POLAR/polartools/issues/89#issuecomment-2867985934, @gfabbris asks:

> Make (or already exists?) a device that just combines a bunch of motors using just kwargs.

prjemian:
- close #1086

@gfabbris: Thanks for the prompt that led to this!


prjemian:
Look at this document to see how to use this new code:
https://github.com/BCDA-APS/apstools/blob/c6e9bd0fdb987a13d7eec0a7f25739edbb342412/docs/source/examples/ho_motor_factory.ipynb

prjemian:
Well, so much for letting VSCode write the type annotations in the unit test file.

prjemian:
I'll add some broadly permissive type annotations to the test file.

gfabbris (approved):
Looks good to me

apstools/devices/motor_factory.py
```diff
@@ -0,0 +1,136 @@
...
+DEFAULT_MOTOR_LABELS: List[str] = ["motors"]
+DEFAULT_KIND: Kind = Kind.hinted | Kind.normal
+
+MOTORS_TYPE = Union[List[str], Dict[str, Optional[Union[str, Dict]]]]
+
+
+def axis_component(parms: dict[str, Any]) -> Component:
+    """Return a Component with 'parms for this axis."""
+    parms = parms or {}
+    if isinstance(parms, str):
```
canismarko, line 49: This doesn't match the type hint. If parms can be a string, then the type hint should be ``str | dict[str, Any]``.

apstools/devices/motor_factory.py
```diff
@@ -0,0 +1,136 @@
...
+        parms = dict(prefix=parms)
+    attrs = dict(kind=DEFAULT_KIND, labels=DEFAULT_MOTOR_LABELS)
+    has_prefix: bool = parms.get("prefix") is not None
```
canismarko, line 52: I think this is mutating the input argument *parms*, right? Might be confusing if client code is not expecting it. Could *parms* be created anew (e.g. ``parms = {**parms}``)? Or else the doc string could make it clear the argument will be mutated?

apstools/devices/motor_factory.py
```diff
@@ -0,0 +1,136 @@
...
+    axis_class_name = parms.pop("class", None)
+    factory = parms.pop("factory", None)
+    if axis_class_name is not None and factory is not None:
+        raise ValueError(f"Unsupported configuration: {axis_class_name=!r} and {factory=!r}")
```
canismarko, line 56: Could this error message describe how the input parameters can be changed to make the configuration supported?

apstools/devices/motor_factory.py
```diff
@@ -0,0 +1,136 @@
...
+
+    args = [base]
+    prefix = attrs.pop("prefix", None)
+    if prefix is not None:
+        args.append(prefix)
+    return Component(*args, **attrs)
+
+
+def mb_class_factory(
+    motors: MOTORS_TYPE = None,
```
canismarko, line 80: I don't think `None` is valid as a `MOTORS_TYPE`. Maybe `MOTORS_TYPE | None`?

apstools/devices/motor_factory.py
```diff
@@ -0,0 +1,136 @@
...
+    class_bases: List[Union[str, Type]] = DEFAULT_DEVICE_BASE_CLASSES,
+    class_name: str = DEFAULT_DEVICE_CLASS_NAME,
+) -> Type:
+    """Create a custom MotorBundle (or as specified in 'class_bases') class."""
+    motors: MOTORS_TYPE = motors or {}
+    if isinstance(motors, list):
+        motors = {axis: {} for axis in motors}
+
+    if not isinstance(class_bases, list):
+        raise TypeError(f"Must be a list, received {class_bases=!r}")
```
canismarko, line 90: This check for classes_base being a list seems unnecessary.

Why must it be a list? Line 95 just needs it to be iterable, and will raise an exception if it's not. Perhaps you can get rid of lines 89-90 and replace the type-hint on line 81 with `collections.abc.Sequence`?

apstools/devices/motor_factory.py
```diff
@@ -0,0 +1,136 @@
...
+    ]
+
+    axes: Dict[str, Component] = {axis: axis_component(parms) for axis, parms in motors.items()}
+    factory_class_attributes: Dict[str, Any] = {}
+    factory_class_attributes.update(**axes)
+
+    def __init__(self, prefix: str = "", **kwargs):
+        for base in class_bases:
+            if hasattr(base, "__init__"):
+                base.__init__(self, prefix=prefix, **kwargs)
```
canismarko, line 105: Why is this explicitly calling `__init__` on class_bases instead of using ``super().__init__``?

apstools/devices/motor_factory.py
```diff
@@ -0,0 +1,136 @@
...
+    class_bases: List[Type] = DEFAULT_DEVICE_BASE_CLASSES,
+    class_name: str = DEFAULT_DEVICE_CLASS_NAME,
+    labels: List[str] = ["motor_device"],
+):
+    """
+    Create MotorBundle with any number of motors.
+    """
+
+    if not isinstance(name, str):
+        raise TypeError("Must provide a 'name', received {name=!r}")
```
canismarko, line 126: Shouldn't *name* just be a required argument then? Why have `None` as a default?

apstools/devices/motor_factory.py
```diff
@@ -0,0 +1,136 @@
...
+
+    if not isinstance(motors, (dict, list)):
```
canismarko, line 128: Why does it need to be a list or dict? What about other things that implement `collections.abc.Mapping` or `collections.abc.Sequence`?

apstools/devices/motor_factory.py
```diff
@@ -0,0 +1,136 @@
...
+
+def mb_creator(
+    *,
+    prefix: str = "",
+    name: Optional[str] = None,
+    motors: MOTORS_TYPE = {},
+    class_bases: List[Type] = DEFAULT_DEVICE_BASE_CLASSES,
+    class_name: str = DEFAULT_DEVICE_CLASS_NAME,
+    labels: List[str] = ["motor_device"],
+):
```
canismarko, line 120: Missing a type hint for the return value? Maybe `): -> bluesky.protocols.HasName`?

apstools/devices/tests/test_motor_factory.py
```diff
@@ -0,0 +1,279 @@
...
+                },
+            },
+            "TestDevice",
+            None,
+            None,
+        ],
+    ],
+)
+def test_axis_component(
+    parms: Dict[str, Any],
```
canismarko, line 78: Parms can also be None (line 31) or str (line 33).

canismarko (changes requested):
I think the basic design is solid. I had done something similar with `type(...)` in Haven and that seemed to be a good approach until ophyd-async.

Many of my comments are around type hints or argument handling, which I think can be cleaned up a bit.

apstools/devices/motor_factory.py
prjemian: Perha[s even `Union[None, str, Dict[str, Any]]`?

apstools/devices/motor_factory.py
prjemian: Good point: Don't mutate the input.  Define a local object from the input.  Apply throughout.  Lather, rinse, repeat.

apstools/devices/motor_factory.py
prjemian: Tried that and ended up with this after some lengthy interaction with LLM guidance.  

Short answer: `super().__init__` failed to do the job right.

apstools/devices/motor_factory.py
prjemian: Matching the signature of any [`OphydObject`](https://github.com/bluesky/ophyd/blob/6410c2b947c6eac9a60d65489f1599f79852100b/ophyd/ophydobj.py#L151) is important for this parameter.  It's a kwarg for OphydObject.  Also, for our guarneri loaders, we only want to be using kwargs.

The reason for `None`, instead of `""` (as I originally had it) is on advice of some static code analysis that says `None` is the safer default value.

```py
    def __init__(self, *, name=None, attr_name="", parent=None, labels=None, kind=None):
```

apstools/devices/motor_factory.py
canismarko: The exception that will be raised on line 95 doesn't cut it?

apstools/devices/motor_factory.py
canismarko: Odd. But okay then.

apstools/devices/motor_factory.py
canismarko: My solution would be `Union[str, Mapping[str, Any]] = {}` and get rid of line 48, but there may be reasons to disfavor the mutable default argument.

apstools/devices/motor_factory.py
prjemian:
```py
In [8]: isinstance("m1", Sequence)
Out[8]: True
```

prjemian: TIL [`typing.Sequence`](https://stackoverflow.com/questions/75285711/difference-between-collections-abc-sequence-and-typing-sequence) is deprecated.

apstools/devices/motor_factory.py
prjemian: exception on line 95?

apstools/devices/motor_factory.py
prjemian: I expanded the typing but left the `None` arg, based on static code review.

apstools/devices/motor_factory.py
prjemian: refactored to avoid mutating the code

apstools/devices/motor_factory.py
prjemian: Revised the exception message

apstools/devices/motor_factory.py
prjemian: done

apstools/devices/motor_factory.py
prjemian: expanded matching class to list, Mapping, set, tuple

apstools/devices/tests/test_motor_factory.py
prjemian: Copied `parms` typing from `motor_factory.py`

prjemian:
@canismarko Recycled your review tag above.  Please check that my changes act on your suggestions.  Thanks!

apstools/devices/motor_factory.py
canismarko: `for base in class_bases` will raise an exception if `class_bases` is not iterable.

Is the `isinstance` check meant to guard against someone passing in e.g. a string, which is iterable but the elements of the sequence are not valid? If that's the case, then don't you actually want to check the elements `class_bases` instead of `class_bases` itself?

apstools/devices/motor_factory.py
canismarko: A string passes `isinstance("m1", Sequence)`, but a list of strings passes `isinstance(['m', '1'], list)` too. Seems like you are actually trying to check the elements of the sequence, not the sequence itself.

It's getting a bit academic, though. I doubt this will cause problems so I'm okay with these checks being left in. Someone might try this with e.g. a generator some day, but they can file that issue when it comes up.

canismarko (approved):
There's an interesting discussion here about whether a check like `isinstance(arg0, list)` is a good idea. I think we've covered it pretty well, and I doubt the edge cases of generators, lists of string, etc will come up often.

I've requested changes because I think `axis_component()` is still mutating its inputs. If that's not the case, then consider this review as "Approve".

canismarko (changes requested):
There's an interesting discussion here about whether a check like `isinstance(arg0, list)` is a good idea. I think we've covered it pretty well, and I doubt the edge cases of generators, lists of string, etc will come up often.

I've requested changes because I think axis_component() is still mutating its inputs. If that's not the case, then consider this review as "Approve".

apstools/devices/motor_factory.py
prjemian: See c5d7cd6

apstools/devices/motor_factory.py
prjemian: I'm imagining someone not getting the YAML quite right for this part.  This check seems friendlier than letting the `dynamic_import()` function, having received the first letter of the one base class provided, declare that a dotted name is required.  The might have provided a dotted name but if it was not a list of str, they'd be very confused.  I'll make the class_base a dotted name in this test case: https://github.com/BCDA-APS/apstools/blob/c5d7cd6c86ccdaad0caf288d8242099015e51438/apstools/devices/tests/test_motor_factory.py#L103-L107

prjemian:
@canismarko How about this handling to avoid mutating the input `parms`?  Very explicit.  Will raise if `parms` is not None, str or some kind of Mapping.

https://github.com/BCDA-APS/apstools/blob/0bfba7bb48e6d07deee9dc714de1ee1ae9372386/apstools/devices/motor_factory.py#L81-L86

canismarko (approved):

prjemian:
Thanks for the reviews!

canismarko:
> Thanks for the reviews!

My pleasure.

PR #1095, opened 2025-05-16
Feature/Add typehints to callbacks and migration.

ravescovi:


prjemian:
This would be easier to review and accept if the changes were confined to the work described in the title.  The additions of code style and JSON file modifications are not necessary to add typehints.

Let's take extremely focused steps as we start to add typehints, starting with only one Python module and its related unit test file.  Style changes, such as the addition of pre-commit should happen in separate branches.

Issue #896, opened 2023-12-24
CI add databroker v2+ to test matrix


prjemian:
Add databroker >1.99 (with tiled >=0.1.0a110) to test matrix.  For databroker ==1.2 (current), use tiled <=0.1.0a96).

With the newer databroker, test with a tiled server.

prjemian:
- refs #643

Issue #835, opened 2023-03-27
new format for catalogs


prjemian:
With databroker 2.0 coming, there will be a new format for catalogs.

## intake: databroker 1.2.5 and previous

This is the current format.

```yml
sources:
  training:
    args:
      asset_registry_db: mongodb://localhost:27017/training-bluesky
      metadatastore_db: mongodb://localhost:27017/training-bluesky
    driver: bluesky-mongo-normalized-catalog
  class_2021_03:
    args:
      asset_registry_db: mongodb://localhost:27017/class_2021_03-bluesky
      metadatastore_db: mongodb://localhost:27017/class_2021_03-bluesky
    driver: bluesky-mongo-normalized-catalog
```

## databroker 2.0.0 uses tiled `profile.yml`

For databroker 2.0.0, it uses a `profile.yml` format.  The long term view is that databroker will
use a tiled server to provide r/w access.  But it is possible to avoid the tiled server by providing
a `direct` configuration.

```yml
training:
    direct:
        trees:
        - path: /
          tree: databroker.mongo_normalized:MongoAdapter.from_uri
          args:
            uri: mongodb://localhost:27017/training-bluesky


class_2021_03:
    direct:
        trees:
        - path: /
          tree: databroker.mongo_normalized:MongoAdapter.from_uri
          args:
            uri: mongodb://localhost:27017/class_2021_03-bluesky
```

## tiled: server `config.yml`

This is a standard tiled `config.yml` for MongoDB catalogs.

```yml
trees:
    - path: training
      tree: databroker.mongo_normalized:Tree.from_uri
      args:
            uri: mongodb://localhost:27017/training-bluesky

  - path: class_2021_03
    tree: databroker.mongo_normalized:Tree.from_uri
    args:
      # for unsecured access
      uri: mongodb://localhost:27017/class_2021_03-bluesky
```

prjemian:
Still use `databroker.catalog_search_path()` to discover list of directories in which to place these files.  The location for databroker 2.0 _will be different_ than before.  Typically `~/.config/tiled/profiles/` is in this list.

prjemian:
See also: https://gist.github.com/danielballan/c37405db4d23c5b05f72a4b4bce9c8fc

prjemian:
- refs #643

Issue #643, opened 2022-02-10
databroker 2.0.0 compels changes in listruns


prjemian:
While testing databroker 2.0.0a22 today, the import failed due to older databroker internals. (attn @danielballan)

```python
In [2]: from tiled.client import from_uri
   ...: c = from_uri("http://127.0.0.1:8000")

In [3]: from databroker.queries import TimeRange, RawMongo
   ...: c["9idc_usaxs"].search(TimeRange(since="2022-02-02", until="2022-02-05")).search(RawMongo(start={"esaf_id": "250815"}))
/share1/bluesky/conda_envs/db2.0/lib/python3.9/site-packages/databroker/queries.py:213: PytzUsageWarning: The zone attribute is specific to pytz's interface; please migrate to a new time zone provider. For more details on how to do so, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html
  timezone = lz.zone
Out[3]: <Catalog {26, 27, 32, 33, 34, 35, 36, 38, 39, 40, ...} ~1317 entries>

In [4]: from apstools.utils import listruns
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-4-15bbcece80f9> in <module>
----> 1 from apstools.utils import listruns

/share1/bluesky/conda_envs/db2.0/lib/python3.9/site-packages/apstools/utils/__init__.py in <module>
----> 1 from .catalog import copy_filtered_catalog
      2 from .catalog import findCatalogsInNamespace
      3 from .catalog import getCatalog
      4 from .catalog import getDatabase
      5 from .catalog import getDefaultCatalog

/share1/bluesky/conda_envs/db2.0/lib/python3.9/site-packages/apstools/utils/catalog.py in <module>
     17 import databroker
     18 import databroker.queries
---> 19 import databroker._drivers.mongo_normalized
     20 import databroker._drivers.msgpack
     21 import logging

ModuleNotFoundError: No module named 'databroker._drivers'

In [5]: databroker.__version__
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-5-c874551783ae> in <module>
----> 1 databroker.__version__

NameError: name 'databroker' is not defined

In [6]: import databroker

In [7]: databroker.__version__
Out[7]: '2.0.0a22'

In [8]: 
```

prjemian:
Look at this after April 2023

danielballan:
[...]

PR #1118, opened 2025-08-22
databroker>2.000

MDecarabas:


prjemian:
Explain in the description box what is the purpose of this PR.

prjemian:
Related issues [still outstanding](https://github.com/BCDA-APS/apstools/issues?q=is%3Aissue%20state%3Aopen%20databroker):

- #643 
- #835 
- #896

prjemian:
The proposed pins a version where `databroker.catalog[catalog_name].v1.insert` is no longer available.  Instruments are expected to load a version of databroker with this method to subscribe to the RunEngine and to obtain APS-specific devices for use in the RunEngine.  Since insert is the current method we have for subscribing to RunEngine documents and storing them in MongoDB, it's a breaking change.  Is there some alternative that would not break current operations?

PR #1124, opened 2025-10-07
1123 specwriter+baseline

### LINKED ISSUE #1123: BUG: specwriter+baseline stream, extra data point written
> I noticed that when we're using the specwriter and the baseline stream, the spec data file contains an extra data point with meaningless information.

> ... when there is a baseline stream, the specwriter reports an extra data point from the primary stream

> that data point has values that are not correct, so it messes up the graphs in pymca

> The way I realized it was baseline interference is because I ran the exact same scan in a session with and without baseline stream. The spec file generated in the session with the baseline has an additional point

prjemian:
- closes #1123

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,7 +24,12 @@ describe future plans.
    1.7.8
    *****
 
-   Release expected by 2025-08.
+   Release expected by 2025-11.
+
+   Fixes
+   -------------
+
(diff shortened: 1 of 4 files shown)


prjemian:
Tests for this PR are passing but other tests (#1126) are failing, blocking this work.  Will resolve those separately and move forward today.

